PR #178, opened 2014-10-05
show user defined variables to the list of completions

kevinb7:
This has a lot of limitations right now.  User variables will only show up in the completion list if they're defined in the top level.  It does, however, not suggest variables before they're defined which is nice.

It uses esprima for the parsing which it does right now inside of the ```getCompletions``` callback.  This is fine for small programs, but will cause problems with longer problems.  The plan to deal with this is do the parsing inside a worker.

TODO list:
- ~~writing tests~~ (half-done, will need to be updated after the other two items are complete)
- ~~do parsing in a web worker~~ (done)
- ~~search for variable declarations in the web worker~~ (done)
- ~~handling variables inside functions and scoping~~ (save for another pull request)
- ~~method completion for methods defined on the prototype~~ (save it for another pull request)

### DIFF
--- a/build-paths.json
+++ b/build-paths.json
@@ -11,7 +11,8 @@
             "external/underscore/underscore.js",
             "external/backbone/backbone.js",
             "external/jed/jed.js",
-            "external/bootstrap/js/bootstrap-tab.js"
+            "external/bootstrap/js/bootstrap-tab.js",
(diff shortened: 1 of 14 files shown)


kevinb7:
I'm going to close this and investigate tern.js instead.  See https://github.com/Khan/live-editor/issues/320 for details.

PR #215, opened 2014-12-22
Make Oh Noes draggable

jdsutton:


jeresig:
@pamelafox @jdsutton  oh rats -- I think we need to make some additional changes to this! having a document ready call in the init function doesn't make sense, and additionally we need to make sure that the draggable plugin exists before being used. This would make a lot more sense if it was integrated directly into the tipbar plugin itself (that way it could be used everywhere). I'll make some changes to it now.

pamelafox:
Ah, I missed that there was a document ready wrapper in there, good catch. I checked third_party_js in webapp and the dependency is setup there (for live-editor on jquery.droppable), but I imagine you're talking about a within-live-editor dependency.

jeresig:
Fixed this up here: f74020d90016c5fbb9e26776c005104046a28324

PR #221, opened 2015-01-07
fix externals/escodgen + remove fullpaths + ES5 debugger support

kevinb7:
The most acute issue was that externals/escodegen was pointing to a ref that didn't exist.  I've updated it to point to the head of origin/master.  I did a fresh clone of kevinb7/live-editor a couple of times to make sure that externals/escodegen was working (it is).  I also noticed that when all externals are checked out they have they're head detached which makes sense.  I probably made a change without first reattaching the HEAD in this repo.

I've iframe-overlay and most of its (and stepper's) dependences to ES6 (using 6to5) and not use fullpaths when browserifying.  You'll notice some changes in how some of the code looks, most of this is due to the switch to 6to5 to build some of the dependencies.  I ran into a hiccup which is 6to5 generates code with 2 space tabs.  The guys over at 6to5 were able to hook me up with an undocumented option to change that.  Right now stepper has 2 space tabs (I'll make that change as a separate pull request) and all of the other dependencies have 4 space tabs.

The ES5 debugger uses regenerator to convert everything by default to ES5... even if ES6 is supported.  You can optionally request native generators.  If they don't exist it will automatically fallback to using regenerator.


build/js/live-editor.debugger.js
```diff
@@ -1,29 +1,44 @@
...
     var paused = false;
-    var queue = new basic.LinkedList();
+    var queue = new LinkedList();
+
     var poster = new Poster(iframe.contentWindow);
+
     function postMouseEvent(e) {
         if (paused) {
-            e["timestamp"] = Date.now();
+            e.timestamp = Date.now(); // Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=238041
```
kevinb7: I used [ ] originally b/c TypeScript doesn't like you adding properties if they aren't part of the type definition.

build/js/live-editor.debugger.js
```diff
@@ -52,67 +67,92 @@ function createOverlay(iframe) {
...
         }
     }
-    overlay.addEventListener("click", function (e) { return postMouseEvent(e); });
-    overlay.addEventListener("dblclick", function (e) { return postMouseEvent(e); });
-    overlay.addEventListener("mouseover", function (e) { return postMouseEvent(e); });
-    overlay.addEventListener("mouseout", function (e) { return postMouseEvent(e); });
+
+    overlay.addEventListener("click", function (e) {
+        return postMouseEvent(e);
+    });
```
kevinb7: The original source was actually: ```overlay.addEventListener("click", e => postMouseEvent(e));```

build/js/live-editor.debugger.js
```diff
@@ -143,9 +185,11 @@ function createRelay(element) {
         });
     });
 }
-exports.createRelay = createRelay;
 
-},{"../node_modules/basic-ds/lib/basic":"/Users/kevin/iframe-overlay/node_modules/basic-ds/lib/basic.js","../node_modules/eventsim/lib/eventsim":"/Users/kevin/iframe-overlay/node_modules/eventsim/lib/eventsim.js","../node_modules/poster/lib/poster":"/Users/kevin/iframe-overlay/node_modules/poster/lib/poster.js"}],"/Users/kevin/iframe-overlay/node_modules/basic-ds/lib/LinkedList.js":[function(require,module,exports){
```
kevinb7: no more absolute paths

build/js/live-editor.debugger.js
```diff
@@ -590,34 +634,8 @@ window.ScratchpadDebugger = Backbone.View.extend({
...
-                "var passed = item.value === 5 && item.done === false;\n" +
-                "item = generator.next();\n" +
-                "passed &= item.value === 6 && item.done === false;\n" +
-                "item = generator.next();\n" +
-                "passed &= item.value === undefined && item.done === true;\n" +
-                "return passed;";
-            return Function(code)();
-        } catch (e) {
-            return false;
-        }
```
kevinb7: we don't need to check if generators are supported anymore because we're using regenerator now  ^_^

demos/simple/output.html
```diff
@@ -16,7 +16,7 @@
     <script>
     var useDebugger = false;
     if (location.search.indexOf("debugger=true") !== -1) {
-        useDebugger = ProcessingDebugger && ProcessingDebugger.isBrowserSupported();
+        useDebugger = typeof ProcessingDebugger === "function";
```
kevinb7: I only caught this when I fired up Safari

external/escodegen
```diff
@@ -1 +1 @@
-Subproject commit ccf0b688b9d60ef4e25c3da961f5fe40073d1924
+Subproject commit 4c3e29bf1ea41f4b0c0d6ccd21b0b0f43f45f6c6
```
kevinb7: It's weird that there were issues with this extern because ```ccf0b688b9d60ef4e25c3da961f5fe40073d1924``` is a legit commit (Merge pull request #1 from kevinb7/yield_with_parentheses).

js/output/pjs/pjs-debugger.js
```diff
@@ -18,7 +19,7 @@ window.PJSDebugger = Backbone.View.extend({
     bind: function() {
         window.addEventListener("message", this.handleMessage.bind(this), false);
 
-        this.debugger.onNewObject = PJSOutput.newCallback.bind(PJSOutput);
+        this.debugger.onNewObject = window.PJSOutput.newCallback.bind(PJSOutput);
```
kevinb7: addresses on of John's nits

PR #231, opened 2015-01-28
Improve page load speed and scrubber responsiveness

kevinb7:
This started out as an experiment to prove some ideas I had to improve page load speed and scrubber responsiveness.  I think the experiment has succeeded.  Here are highlights of the results:

- core_deps is 50% of what it used to be, output_pjs_deps is 30% of what it used to be
- uncached page load was 1500~2500ms and is now 1000ms~1300ms
- cached page load was 1200~1500ms and is now 900ms~1100ms
- memory usage was 24.9 MB (main heap) – 37.3 MB (total) and is now 19.8 MB (main heap) – 33.5 MB (total)
- scrubbing performance was ~15 fps and is now ~35 fps

Note: all numbers are for live-editor only.

For more details and live demos check out http://kevinb7.github.io/live-editor/.


.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "external/debugger"]
 	path = external/debugger
 	url = https://github.com/kevinb7/debugger
+[submodule "external/p5"]
+	path = external/p5
+	url = https://github.com/kevinb7/p5.js
```
kevinb7: p5.js is quite a bit smaller than processing-js and is more modular.  I've made some pull requests (which have been accepted) to add a few things to make it easier to integrate.  There's one bug that needs fixing https://github.com/lmccart/p5.js/issues/492 and in general it needs more testing.  I'd like to see an image compare test suite similar to what processing-js has.

build-paths.json
```diff
@@ -6,13 +6,17 @@
         "core_deps": [
             "bower_components/es5-shim/es5-shim.js",
             "bower_components/jquery/dist/jquery.js",
-            "bower_components/jquery-ui/ui/jquery-ui.custom.js",
+            "bower_components/jquery-ui/ui/jquery.ui.core.js",
+            "bower_components/jquery-ui/ui/jquery.ui.widget.js",
+            "bower_components/jquery-ui/ui/jquery.ui.mouse.js",
+            "bower_components/jquery-ui/ui/jquery.ui.draggable.js",
```
kevinb7: replace jquery-ui.custom.js which was big with just the modules being used

bower.json
```diff
@@ -27,7 +27,8 @@
     "jed": "0.5.0",
     "escope": "~1.0.0",
     "iframe-overlay": "0.1.4",
-    "poster": "0.1.0"
+    "poster": "0.1.0",
+    "dialog-polyfill": "~0.2.0"
```
kevinb7: only Chrome supports the <dialog> tag so let's polyfill it (it's smaller than jquery-ui's modal)

build/css/live-editor.ui.css
```diff
@@ -390,35 +390,35 @@ button.hilite {
...
+  color: black;
 
-/* ace_cursor */
+  width: -moz-fit-content;
+  width: -webkit-fit-content;
+  width: fit-content;
 
-.debugging .ace_cursor {
-    display: none;
-}
```
kevinb7: debugging styles are in their own .css now

build/js/live-editor.editor_ace.js
```diff
@@ -47,19 +47,25 @@ window.AceEditor = Backbone.View.extend({
...
 
-        // TODO(bbondy): Support multiple content types for autosuggest.
-        if (this.tooltips[this.type].indexOf("autoSuggest") !== -1) {
-            ScratchpadAutosuggest.init(this.editor);
-        }
+            // TODO(bbondy): Support multiple content types for autosuggest.
+            if (this.tooltips[this.type].indexOf("autoSuggest") !== -1) {
+                ScratchpadAutosuggest.init(this.editor);
+            }
+        }.bind(this), 500);
```
kevinb7: defer TooltipEngine setup so that the page can render faster

build/js/live-editor.output.js
```diff
@@ -6,11 +6,17 @@ this["Handlebars"]["templates"]["output"] = Handlebars.template(function (Handle
 
 
   return "<div class=\"output\"></div>\n<div class=\"test-errors\" style=\"display: none;\"></div>";});;
-var PooledWorker = function(filename, onExec) {
+var PooledWorker = function(url, onExec) {
```
kevinb7: switch to using URLs so that we can passing Blob object URLs (or data URIs)

build/js/live-editor.output.js
```diff
@@ -117,12 +123,21 @@ OutputTester.prototype = {
...
+                        errors: errors,
+                    });                    
+                } else {
+                    worker.postMessage({
+                        code: code,
+                        validate: validate,
+                        errors: errors,
+                        deps: options.deps
+                    });
+                    worker.initialized = true;
```
kevinb7: only send dependencies to the worker on the first postMessage

build/js/live-editor.output_pjs.js
```diff
@@ -1213,14 +1215,29 @@ window.PJSOutput = Backbone.View.extend({
...
+        if (window.localStorage.imageFilenames) {
+            imageFilenames = JSON.parse(window.localStorage.imageFilenames);
+        } else {
+            imageFilenames = this.getImageFilenames(options.code);
+        }
+        this.cacheImages(imageFilenames, function () {
+            this.injectCode(options.code, function(errors) {
+                console.log(errors);
+            });
+        }.bind(this));
```
kevinb7: Start the image caching process before first run.  This could be improved by creating a promise and then waiting for the promise to resolve before attempting to run.

build/js/live-editor.output_pjs.js
```diff
@@ -1322,12 +1339,186 @@ window.PJSOutput = Backbone.View.extend({
...
+                "es5-shim.js": data["es5-shim.js"],
+                "jshint.js": data["jshint.js"],
+                "underscore.js": data["underscore.js"]
+            };
+
+            // TODO(kevinb7) polyfill vendor prefixed version to this
+            // TODO(kevinb7) fallback to data URI on IE9 (or just use a real URL)
+            var jshintBlob = new Blob([data["jshint-worker.js"]], { type: 'application/javascript' });
+            this.hintWorker = new PooledWorker(
+                URL.createObjectURL(jshintBlob),
```
kevinb7: The workers themselves are included in deps.json.  createObjectURL can create page-specific? URLs from Blobs and Files.  This needs to be made more robust see lines 1358 and 1359 for TODOs.

build/js/live-editor.output_pjs.js
```diff
@@ -1322,12 +1339,186 @@ window.PJSOutput = Backbone.View.extend({
...
+
+        // TODO(kevinb7) change this.canvas to something like this.p5
+        this.canvas = new p5(function (instance) {
+            instance.draw = this.DUMMY;
+            instance.setup = this.DUMMY;
+        }.bind(this), this.$el[0], true);   // true = async
+        
+        // monkey patch size so that it doesn't throw 'size() not implemented...'
+        this.canvas.__loadImage = this.canvas.loadImage;
+        this.canvas.size = this.canvas.resizeCanvas;
```
kevinb7: p5.js doesn't have a "size" method

build/js/live-editor.output_pjs.js
```diff
@@ -1322,12 +1339,186 @@ window.PJSOutput = Backbone.View.extend({
...
+        
+        // using bindProcessing doesn't work because it binds the "this" value 
+        // in functions which we don't want to because p5.Vector is a constructor
+        this.canvas.PVector = p5.Vector;
+
+        // TODO(kevinb7) both map to console.log, need to create virtual console
+        // Note: needed to wrap it here because tests check function.length
+        this.canvas.println = function(msg) {
+            console.log(msg);
+        };
```
kevinb7: Need to implement a virtual console.

build/js/live-editor.output_pjs.js
```diff
@@ -1322,12 +1339,186 @@ window.PJSOutput = Backbone.View.extend({
...
+            var jshintBlob = new Blob([data["jshint-worker.js"]], { type: 'application/javascript' });
+            this.hintWorker = new PooledWorker(
+                URL.createObjectURL(jshintBlob),
+                function(hintCode, callback) {
+                    // Fallback in case of no worker support
+                    if (!window.Worker) {
+                        if (!JSHINT) {
+                            eval(data["jshint.js"]);
+                            window.JSHINT = JSHINT;
+                        }
```
kevinb7: only load JSHINT if we actually need it, saves memory use by 0.9 MB.

build/js/live-editor.output_pjs.js
```diff
@@ -1940,44 +2106,6 @@ window.PJSOutput = Backbone.View.extend({
 
             var context = {};
 
-
-            // We can send object literals over, but not
-            //  objects created with a constructor.
-            // jQuery thinks PImage is a plain object,
-            //  so we must specially check for it,
-            //  otherwise we'll give web workers an object that
-            //  they can't serialize.
```
kevinb7: p5.Image doesn't seem to need any special checks.

build/js/live-editor.output_pjs.js
```diff
@@ -2369,7 +2495,7 @@ window.PJSOutput = Backbone.View.extend({
             this.clear();
 
             // Clear Processing logs
-            this.canvas._clearLogs();
```
kevinb7: p5.js doesn't have a _clearLogs method... will need to add this back in when implementing the virtual console

build/js/live-editor.tooltips.js
```diff
@@ -2134,6 +2134,7 @@ TooltipEngine.classes.colorPicker = TooltipBase.extend({
                 var $picker = $(this);
                 $picker.addClass("active");
                 down = true;
+                self.options.editor.scrubberActive = true;
```
kevinb7: it might be better to use events to communicate changes in scrubber status to the outside world

build/js/live-editor.tooltips.js
```diff
@@ -2362,7 +2371,7 @@ TooltipEngine.classes.colorPicker = TooltipBase.extend({
                 soundsDir: this.options.soundsDir,
                 classes: this.options.files
             }));
-            this.$el.appendTo("body").hide();
+            this.$el.appendTo("body");
```
kevinb7: $el[0] is now a <dialog> element which is hidden by default

build/js/live-editor.tooltips.js
```diff
@@ -2852,6 +2865,13 @@ TooltipEngine.classes.numberScrubber = TooltipBase.extend({
                     setTimeout(function () {
                         self.dragged = false;
                     }, 0);
+
+                    self.parent.editor.scrubberActive = false;
+                    
+                    // updateText so that live-editor.js:runCode can signal pjs-output.js to stop profiling
+                    self.intermediateValue += Math.pow(10,-self.decimals);
+                    self.updateText(self.intermediateValue.toFixed(self.decimals));
+                    self.dragged = true;
```
kevinb7: It might be better to use events to communicate changes in scrubber status to live-editor.js

build/js/live-editor.ui.js
```diff
@@ -973,15 +948,25 @@ window.LiveEditor = Backbone.View.extend({
...
-        this.$el.html(Handlebars.templates["live-editor"]({
-            execFile: this.execFile,
-            imagesDir: this.imagesDir,
-            colors: this.colors
-        }));
+        //this.$el.html(Handlebars.templates["live-editor"]({
+        //    execFile: this.execFile,
+        //    imagesDir: this.imagesDir,
+        //    colors: this.colors
+        //}));
```
kevinb7: don't render anymore because the HTML for this template is hard coded in index.html.  this was done so that we could start loading output.html sooner.

build/workers/pjs/jshint-worker.js
```diff
@@ -1,30 +1,37 @@
...
-        importScripts(event.data.externalsDir +
-            "es5-shim/es5-shim.js?cachebust=" + (new Date()).toDateString());
+        if (data.deps) {
+            var deps = data.deps;
 
-        importScripts(event.data.jshintFile +
-            "?cachebust=" + (new Date()).toDateString());
+            eval(deps["es5-shim.js"]);
+            eval(deps["jshint.js"]);
+            eval(deps["underscore.js"]);
```
kevinb7: load dependencies using eval

build/workers/pjs/jshint-worker.js
```diff
@@ -1,30 +1,37 @@
...
+    
+    var hintData = JSHINT.data();
+    var globals = hintData.globals || [];
 
     // Return the JSHint results to the main code
     self.postMessage({
          type: "jshint",
          message: {
-             hintData: JSON.parse(JSON.stringify(JSHINT.data())),
+             hintData: { globals: globals },
```
kevinb7: only send globals because that's all we use

build/workers/pjs/worker.js
```diff
@@ -48,7 +59,11 @@ self.onmessage = function(event) {
                         stubbedContext[prop] :
                         drawCounter(prop);
                 } else if (typeof object[prop] === "object") {
-                    unstubFunctionsInObject(object[prop]);
+                    if (prop === "Program" && prop in stubbedContext) {
+                        object[prop] = stubbedContext[prop];
```
kevinb7: for some reason "Program" was being unstubbed

demos/simple/index.html
```diff
@@ -23,7 +23,104 @@
...
+                <div class="scratchpad-toolbar">
+                    <button id="restart-code"
+                            class="simple-button pull-right">
+                        <span class="icon-refresh"></span>
+                        Restart
+                    </button>
+
+                    <!-- Widgets for selecting colors to doodle on the canvas during
+                        recordings -->
+                    <!--<div id="draw-widgets" style="display:none;">-->
```
kevinb7: I commented out the stuff for making recordings.  It would be nice if this HTML (and associated JavaScript) was dynamically loaded only when making recordings.

demos/simple/index.html
```diff
@@ -32,15 +129,22 @@
     <script src="../../build/js/live-editor.tooltips.js"></script>
     <script src="../../build/js/live-editor.ui.js"></script>
     <script src="../../build/js/live-editor.editor_ace.js"></script>
-    <script src="../../build/js/live-editor.debugger.js"></script>
+    <!--<script src="../../build/js/live-editor.debugger.js"></script>-->
```
kevinb7: need to conditionally load this file.  also, need to optimize its dependencies.

demos/simple/output.html
```diff
@@ -7,12 +7,13 @@
...
 
-    <script src="../../build/js/live-editor.core_deps.js"></script>
+    <!-- TODO(kevinb7) use JSON dependency loading system for core_deps.js and shared.js -->
+    <script src="../../build/js/live-editor.core_deps.js?cachebust=1"></script>
     <script src="../../build/js/live-editor.shared.js"></script>
     <script src="../../build/js/live-editor.output_pjs_deps.js"></script>
     <script src="../../build/js/live-editor.output.js"></script>
     <script src="../../build/js/live-editor.output_pjs.js"></script>
-    <script src="../../build/js/live-editor.output_debugger_deps.js"></script>
+    <!--<script src="../../build/js/live-editor.output_debugger_deps.js"></script>-->
```
kevinb7: need to optimize the dependencies so that we're sharing esprima.js which is loaded as part of deps.json.

demos/simple/index.html
kevinb7: I thought about this some more and it doesn't really fit with the rest of what's going on in this pull request.  I'll uncomment this.  In the future though I would like to separate out video playback and recording code/markup so that it can be loaded only when it's needed.

build/workers/pjs/jshint-worker.js
```diff
@@ -1,32 +0,0 @@
-var init = false;
```
kevinb7: removed all files in build/workers/pjs because all of these files are being packaged in deps.json

js/workers/pjs/jshint-worker.js
```diff
@@ -1,30 +1,37 @@
...
-        importScripts(event.data.externalsDir +
-            "es5-shim/es5-shim.js?cachebust=" + (new Date()).toDateString());
+        if (data.deps) {
+            var deps = data.deps;
 
-        importScripts(event.data.jshintFile +
-            "?cachebust=" + (new Date()).toDateString());
+            eval(deps["es5-shim.js"]);
+            eval(deps["jshint.js"]);
+            eval(deps["underscore.js"]);
```
kevinb7: use eval to load dependencies instead of importScripts

package.json
```diff
@@ -17,24 +17,26 @@
...
-    "run-sequence": "^0.3.6",
-    "async": "^0.9.0",
+    "q": "^1.1.2",
     "request": "^2.44.0",
-    "sinon": "^1.10.3",
-    "handlebars": "1.0.5-beta",
-    "gulp-bower": "0.0.7",
-    "gulp-symlink": "^2.0.1"
+    "run-sequence": "^0.3.6",
+    "sinon": "^1.10.3"
```
kevinb7: That's what I get for calling ```npm install q --save-dev```.

tests/output/pjs/output_test.js
```diff
@@ -202,7 +202,7 @@ describe("Scratchpad Output Exec", function() {
 
     test("createFont Inside Sub-Object", function() {
         var test = [{
-            test: createFont("fantasy", 20)
+            test: "fantasy"
```
kevinb7: p5.js doesn't have a createFont()

tests/output/pjs/test_utils.js
```diff
@@ -122,9 +122,11 @@ var runTest = function(options) {
                     checkAssertions(options.assertions2,
                         output.results.assertions);
                     options.simulateClick && simulateClick();
+                    output.output.kill();
                     done();
                 });
             } else {
+                output.output.kill();
```
kevinb7: Cleanup before running the next test.  kill() now calls noLoop() which clears the interval that p5.js uses for ```draw```.  This resulted in a significant reduction in time to run the tests.  They used to take about a 90 seconds for me, now they take about 45 seconds.

.gitmodules
pamelafox: It worries me a bit to move to P5.JS without doing a lot of testing, and without having an official team to tackle issues (I'm currently the only eng on CS, and I'm creating content this week). One thing I could do is deploy this change with p5.JS to a non-default server, and have you poke at a bunch of programs and watch talk-throughs.

build-paths.json
pamelafox: Please make sure ?blocks still works (structured-blocks), as that was a big user of jQuery.UI parts.

build/css/live-editor.ui.css
```diff
@@ -390,35 +390,35 @@ button.hilite {
     box-shadow: 0 0 20px green;
 }
 
-/*
- * Contains styles for breakpoints and modifies the editor's appearance
- * when debugging.
- */
+dialog {
```
pamelafox: The modal change is also something that I'd want heavy testing of. We had to make several modal changes in December when we got feedback about modals on small screens, ChromeBooks, etc.

demos/simple/index.html
```diff
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="../../build/css/live-editor.audio.css"/>
     <link rel="stylesheet" href="../../build/css/live-editor.tooltips.css"/>
     <link rel="stylesheet" href="../../build/css/live-editor.ui.css"/>
-    <link rel="stylesheet" href="../../build/css/live-editor.debugger.css"/>
+    <!--<link rel="stylesheet" href="../../build/css/live-editor.debugger.css"/>-->
```
pamelafox: Commented out?

demos/simple/index.html
```diff
@@ -23,7 +23,105 @@
 </head>
 <body>
     <h1>Live Editor Example</h1>
-    <div id="sample-live-editor"></div>
+    <div id="sample-live-editor">
```
pamelafox: Did you copy the HTML from khanacademy.org? (I'm just curious)

.gitmodules
kevinb7: For sure.  This is a huge change.  I found a bug last relating to how it implements "color".  This whole pull request is more an experiment.  I made a it a pull request so you and the team could see various changes that we may want to do in the future.  I started out actually pull stuff out of processing-js like the Java parser and all the 3D code and then I though it might be easier to integrate p5.js (and you had mentioned something about doing that a while ago) so I thought I'd give it a try.

demos/simple/output.html
```diff
@@ -7,12 +7,12 @@
 <body>
     <div id="live-editor-output"></div>
 
-    <script src="../../build/js/live-editor.core_deps.js"></script>
+    <!-- TODO(kevinb7) use JSON dependency loading system for core_deps.js and shared.js -->
```
pamelafox: Might be best to use whatever Perseus uses for their local demo, if we move to that here. Or at least check it out to see what that repo uses.

demos/simple/output.html
```diff
@@ -7,12 +7,12 @@
...
+    <script src="../../build/js/live-editor.core_deps.js?cachebust=1"></script>
```
pamelafox: Probably can remove cache_bust now?

gulpfile.js
```diff
@@ -64,8 +65,8 @@ gulp.task("workers", function() {
     gulp.src(paths.workers_webpage)
         .pipe(gulp.dest("build/workers/webpage"));
 
-    gulp.src(paths.workers_pjs)
-        .pipe(gulp.dest("build/workers/pjs"));
+    // TODO(kevinb7) make this more gulp-y
```
pamelafox: Not sure what more gulp-y means

js/editors/ace/editor-ace.js
```diff
@@ -47,19 +47,25 @@ window.AceEditor = Backbone.View.extend({
...
-            type: this.type,
-            imagesDir: options.imagesDir,
-            soundsDir: options.soundsDir,
-            editor: this.editor,
-            record: this.record
-        });
+        // defer this until after the first run of output
+        // if the first run takes longer than 500ms then 
+        // start the tooltip engine anyways
+        // this saves about 100 ms during start
```
pamelafox: Nice idea

js/live-editor.js
```diff
@@ -1,4 +1,6 @@
 window.LiveEditor = Backbone.View.extend({
+    el: '#sample-live-editor',
```
pamelafox: Hm, I don't know that we want an ID hard-coded here?

js/live-editor.js
```diff
@@ -65,37 +66,8 @@ window.LiveEditor = Backbone.View.extend({
...
-
-                // If a new color has actually been chosen
-                if (color !== null) {
-                    // Select that color and activate the button
-                    this.$el.find("#" + color).addClass("ui-state-active");
-                }
-            }.bind(this)
-        });
+        // should only do this if there's a flag set
+        //this.setupDrawCanvas();
```
pamelafox: Commented out?

js/live-editor.js
```diff
@@ -166,15 +138,25 @@ window.LiveEditor = Backbone.View.extend({
         }
 
         this.bind();
+
+        // jquery.ui.slider.js is require for the toolbar and audio to work
```
pamelafox: Commented out?

js/live-editor.js
```diff
@@ -166,15 +138,25 @@ window.LiveEditor = Backbone.View.extend({
...
     },
 
     render: function() {
-        this.$el.html(Handlebars.templates["live-editor"]({
-            execFile: this.execFile,
-            imagesDir: this.imagesDir,
-            colors: this.colors
-        }));
+        //this.$el.html(Handlebars.templates["live-editor"]({
+        //    execFile: this.execFile,
```
pamelafox: Commented out?

js/live-editor.js
```diff
@@ -195,9 +177,11 @@ window.LiveEditor = Backbone.View.extend({
         this.handleMessagesBound = this.handleMessages.bind(this);
         $(window).on("message", this.handleMessagesBound);
 
+        // the output frame is getting the code from the server and it knows
+        // whether it's good to run or not
         $el.find("#output-frame").on("load", function() {
             this.outputState = "clean";
-            this.markDirty();
+            //this.markDirty();
```
pamelafox: Commented out?

pamelafox:
Can you  list and rank the big changes made by their perf benefits? And maybe make separate pull requests for the small ones that are lower risk?

kevinb7:
I'll do that after lunch.  Thanks for looking this over.  Smaller pull requests are definitely better.

build-paths.json
kevinb7: TL;DR: I won't be pursing this because I'd rather have a system that handles minimizing dependencies in a more automatic fashion.  See below for details.

I was thinking that it might be better to make more tailored packages for each of the different environments.  This would make each of download smaller, but if a user is switching between multiple environments this would make the downloads big.  Part of the issue with this is that the script tags have to be executed in a particular order so we can't just have a true shared.js and then a separate environment.js because some of the shared dependencies need to be loaded later.  What I'd like to do is the JSON dependency loading system that's used for the workers, but modify it so that to use "require".  I have a prototype of this working using a custom implementation of "require" so it is possible.  Using "require" would allow us to only load the dependencies that we're actually using which would cut down our memory foot print.  Also, we could add a hook to our custom "require" implementation to trace which dependencies are used by each environment.  From this we could determine a minimal "shared.js" and a minimal "pjs_deps.js", "sql_deps.js", "web_deps.js", etc.

build/css/live-editor.ui.css
kevinb7: Of course.  What kind of issues did you see?  How small is small?  Also, it seems like there's a different modal that's used for some things in production in addition the jquery-ui modal that's used by the live-editor itself for the picker.  It would be nice if there was some way to use the same modal code.

demos/simple/index.html
kevinb7: This should be removed.  I added code to conditionally load the JavaScript and CSS for debugger only if the user as added ```?debugger=true```.

demos/simple/index.html
kevinb7: I used the code from ```live-editor.handlebars``` and remove the templating directives/macros.

demos/simple/output.html
kevinb7: Will do.

demos/simple/output.html
kevinb7: I would've thought so too.  Chrome was being quite silly.  It may have also been the local server that I was using.  It was really weird... Chrome would load the correct version in index.html and then load an old cached one in output.html.

gulpfile.js
kevinb7: gulp is all about processing streams and ```bundle_deps.bundle_deps(bundle_deps.deps);``` (on the following line) doesn't really do that.  All of the dependencies are listed directly in bundle_deps.js.  It would be nice if those paths could be passed in via a glob and then have it output a stream so that additional processing could be done like minification.

js/live-editor.js
kevinb7: I can set the element in ```initialize``` instead.

js/live-editor.js
kevinb7: I was trying to see how quickly I could get the page to load for the case where a student is working on a program in the live-editor.  There are additional pieces of code that are packaged with live-editor but aren't always used: playback code used when students watch a lesson and recording code used when you create a lesson.  I was thinking that page load speeds could be improved if we only load these components when they're required.  Not loading these components also improves memory use.

js/live-editor.js
kevinb7: See my previous comment.

js/live-editor.js
kevinb7: This should just be removed now that the HTML for live-editor is directly in index.html.  I'm curious though about how live-editor gets packaged for production.  The demo/simple/index.html isn't used in production, is it?  If that's the case, we'd have to make sure that the HTML that's in index.html actually makes its way into production.

js/live-editor.js
kevinb7: This line can be removed.  This initial ```markDirty()``` is no longer required because output.html starts running code independently of index.html.  It's only when the user starts making changes to index.html that ```markDirty()``` comes into play.

kevinb7:
Changes:
- skip linting when using the scrubber
  - benefit: big improvement in scrubber responsiveness (increase from 15fps to 35fps)
  - effort: low
- load web workers and their dependencies from a JSON blob
  - benefit: decrease the number of network calls, biggest impact seen with slow networks
  - effort: low
- load JSHINT from JSON blob if web workers are unavailable
  - benefit: decrease memory use, reduce the size of output_pjs_deps
  - effort: low
- split up jquery-ui.custom.js
  - benefit: reduces core_deps.js by 50%, improves load speed and memory use
  - effort: medium, requires separate bundles for each environment for maximum effect, need to test dialog-polyfill as a replacement for jquery modal.
- running user code from server in output.html immediately
  - benefit: user's program starts much faster
  - effort: medium, requires server changes to load user code directly into output.html, should also store whether the program is safe to run, i.e. it was linted and passed all of the tests
- replace processing-js with p5.js
  - benefit: much smaller code base that processing-js, more modular, has plugins
  - effort: high, need to fix bugs in p5.js (color bug, arc bug, probably more), not the same level of maturity , no visual comparison tests
- making processing-js more modular
  - benefits: don't include code that isn't being used, e.g. Java parser, 3D code
  - effort: medium, processing-js has a good test suite so as long as that keeps passing we should be able to refactor to our hearts' content
- make a minimal shared bundle and environment specific bundles using JSON dependencies and a custom "require" implementation:
  - benefits: reduce the size of bundles and the amount of JS code that's being loaded, automate dependency management
  - effort: high, need to write tests for custom "require" implementation, need to write code determine shared dependencies between environments, need to update code to use "require" statements instead of using globals

pamelafox:
Great summary!

There are a few that I would avoid for now because of the way they interact with our internal repo/app:
* running user code from server in output.html immediately
* split up jquery-ui.custom.js/custom require- we have an internal bundling system and I would want to consult with that team before adding a require system to live-editor. There's also an interaction between the 2.

The first three sound like they'd be good individual pull requests, and nicely limited in scope to the live-editor repo.

Re P5: I'd be happy to deploy a server with P5 on it to make it easier to poke around at it, but I don't think it's worth it until all the known bugs are gone. But I do suspect it's easier to refactor processing-js --we've made a decent number of custom modifications to it. One of the main features I like in p5 is the retina display support, but I think we could bring that into our processing-JS.

kevinb7:
Sounds good.  I'm going to close this pull request then and I'll start with the first three (one at at time).  I'll continue to work on fixing the outstanding issues that I know of in p5.

PR #240, opened 2015-02-05
skip lint when scrubbing using the color picker or the number scrubber

kevinb7:


js/live-editor.js
```diff
@@ -111,6 +111,17 @@ window.LiveEditor = Backbone.View.extend({
             workersDir: this.workersDir,
             type: this.editorType
         });
+        
+        this.noLint = false;
+        this.editor.on("scrubbingStarted", function() {
+            console.log("scrubbingStarted");
```
pamelafox: Do we need the console.log()s?

kevinb7: nope... thanks for catching that.

pamelafox:
Great, merging!

Issue #241, opened 2015-02-05
number scrubber arrows aren't aligned


kevinb7:
Observed:
![screen shot 2015-02-05 at 10 11 27 am](https://cloud.githubusercontent.com/assets/1044413/6065079/acaa3b42-ad1f-11e4-8a1e-b47999ce1e49.png)

Desired:
![screen shot 2015-02-05 at 10 11 15 am](https://cloud.githubusercontent.com/assets/1044413/6065081/afd6d136-ad1f-11e4-8de2-59aa3bd6841a.png)

PR #242, opened 2015-02-05
Correct alignment of right arrow on number scrubber

kevinb7:
Addresses issue https://github.com/Khan/live-editor/issues/241.

### DIFF
--- a/css/ui/tooltips.css
+++ b/css/ui/tooltips.css
@@ -399,8 +399,11 @@ html.playing .tooltip.hide-while-playing {
 
 .scrubber-handle .flipped-arrow {
     display: inline-block;
-    transform: rotate(180deg);
-    -webkit-transform: rotate(180deg);
+    transform: scaleX(-1.0);
+    -webkit-transform: scaleX(-1.0);
+    -moz-transform: scaleX(-1.0);
+    -ms-transform: scaleX(-1.0);
+    -o-transform: scaleX(-1.0);
 }
 
 .block .tooltip:not(.picker) {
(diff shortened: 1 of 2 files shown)


css/ui/tooltips.css
```diff
@@ -399,8 +399,11 @@ html.playing .tooltip.hide-while-playing {
 
 .scrubber-handle .flipped-arrow {
     display: inline-block;
-    transform: rotate(180deg);
-    -webkit-transform: rotate(180deg);
+    transform: scaleX(-1.0);
+    -webkit-transform: scaleX(-1.0);
+    -moz-transform: scaleX(-1.0);
+    -ms-transform: scaleX(-1.0);
+    -o-transform: scaleX(-1.0);
```
kevinb7: I'd like to suggest migrating to stylus + nib for doing CSS.  nib provides auto-prefixing and stylus could simplify our nested rules and provide nice things like variables for common colors and stuff.

http://learnboost.github.io/stylus/
http://nibstyl.us/docs/#transparent-mixins

pamelafox: We use LESS internally at KA, so that's what we'd want to move to if we migrated this repo over.

kevinb7: I found this auto-prefixer for LESS: http://lessprefixer.com/.  Is this what other repos are using for auto-prefixing?

kevinb7: I also found https://github.com/postcss/autoprefixer which doesn't require either LESS (or stylus).  It also appears to be much better maintained than lessprefixer.

pamelafox: Hm. It's possible we don't actually have an autoprefixing system in place internally. I'm poking around.

pamelafox: Aha, we use https://github.com/postcss/autoprefixer, it gets auto-run over our code.

Issue #223, opened 2015-01-09
If sound hasn't loaded and is played, the program crashes


MikaalSky:
Example: https://www.khanacademy.org/computer-programming/target-shooting-sound/5043290630782976

Click immediately after loading and keep clicking. The UI will half disappear.

pamelafox:
Can you open your Chrome Dev Tools when it happens and tell me what the error is? I haven't managed to click it fast enough yet.

MikaalSky:
InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable scratchpads-exec-pjs-package-ddf410.js:7287

kevinb7:
We probably want to do something similar to how images are loaded where there's callbacks for ```onload``` and ```onerror```.  HTMLAudioElements don't have respond to ```onload```.  Here's a link to an MDN page with a list of the events: https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Media_events.  We probably want ```oncanplaythrough```.

kevinb7:
I haven't been able to reproduce this error.  I tried to recreate it in isolation with a small example that sets audioElement.src and then calls audioElement.play() synchronously.  It appears that Chrome (	41.0.2272.35 on Mac OS X) queues ```play()``` calls and starts playing after the file has loaded.  I even tried simulating poor network connectivity to see if that would cause the InvalidStateError, no luck.

I did some Googling about InvalidStateError.  This issue usually occurs when the element isn't ready  for whatever reason.  People suggest using the following pattern to deal with the issue:
```
if (domObject.readyState > 0){
    // do something with domObject
  }
else{
    domObject.onload = function(e){  // or "onopen" or "oncanplaythrough"... whatever makes sense
        // do something with domObject
    }
}
```
Since we can attach the ```oncanplaythrough``` handler before we set the ```src``` the above pattern can be simplified.

TL;DR: I can't repro the bug, but I'm going to add an ```oncanplaythrough``` handler and do something similar to how images are loaded.

kevinb7:
Fixed by https://github.com/Khan/live-editor/pull/243.

PR #243, opened 2015-02-05
Cache sounds before running user code

kevinb7:
Attempt to address https://github.com/Khan/live-editor/issues/223.

### DIFF
--- a/build/js/live-editor.output.js
+++ b/build/js/live-editor.output.js
@@ -529,11 +529,13 @@ window.LiveEditorOutput = Backbone.View.extend({
             // Normal case
             } else {
                 // This is debounced (async)
-                this.test(userCode, this.validate, errors, function(errors, testResults) {
-                    this.results.errors = errors;
-                    this.results.tests = testResults;
-                    this.phoneHome();
-                }.bind(this));
+                if (this.validate !== "") {
+                    this.test(userCode, this.validate, errors, function(errors, testResults) {
+                        this.results.errors = errors;
+                        this.results.tests = testResults;
+                        this.phoneHome();
+                    }.bind(this));   
+                }
             }
         }.bind(this);
 
--- a/build/js/live-editor.output_pjs.js
+++ b/build/js/live-editor.output_pjs.js
@@ -1530,8 +1530,7 @@ window.PJSOutput = Backbone.View.extend({
     },
 
     imageCache: {},
-    imagesCached: false,
-    imageCacheStarted: false,
+    soundCache: {},
     imageHolder: null,
 
     // Load and cache all images that could be used in the environment
@@ -1540,9 +1539,10 @@ window.PJSOutput = Backbone.View.extend({
     // Execution is delayed once a getImage appears in the source code
     // and none of the images are cached. Execution begins once all the
     // images have loaded.
-    cacheImages: function(userCode, callback) {
+    cacheResources: function(userCode, callback) {
         // Grab all the image calls from the source code
         var images = userCode.match(/getImage\s*\(.*?\)/g);
+        var sounds = userCode.match(/getSound\s*\(.*?\)/g);
 
         // Keep track of how many images have loaded
         var numLoaded = 0;
@@ -1559,13 +1559,21 @@ window.PJSOutput = Backbone.View.extend({
                 })
                 .appendTo("body");
         }
+        
+        var resourceCount = 0;
+        if (images !== null) {
+            resourceCount += images.length;
+        }
+        if (sounds !== null) {
+            resourceCount += sounds.length;
+        }
 
         // Keep track of when image files are loaded
         var loaded = function() {
             numLoaded += 1;
 
             // All the images have loaded so now execution can begin
-            if (numLoaded === images.length) {
+            if (numLoaded === resourceCount) {
                 callback();
             }
         };
@@ -1577,12 +1585,16 @@ window.PJSOutput = Backbone.View.extend({
 
             // Skip if the image has already been cached
             // Or if the getImage call is malformed somehow
-            if (this.imageCache[file] || !fileMatch) {
+            if (!fileMatch) {
                 return loaded();
             }
-
+            
             file = fileMatch[1];
-
+            
+            if (this.imageCache[file]) {
+                return loaded();
+            }
+            
             // We only allow images from within a certain path
             var path = this.output.imagesDir + file + ".png";
 
@@ -1602,10 +1614,38 @@ window.PJSOutput = Backbone.View.extend({
             // but PImage may be mutable, so that might not work.
             this.imageCache[file] = img;
         }.bind(this));
+        
+        _.each(sounds, function(file) {
+            var fileMatch = /["']([A-Za-z0-9_\/-]*?)["']/.exec(file);
+
+            if (!fileMatch) {
+                return loaded();
+            }
+
+            file = fileMatch[1];
+            
+            if (this.soundCache[file]) {
+                return loaded();
+            }
+            
+            var audio = document.createElement("audio");
+            
+            audio.preload = "auto";
+            audio.oncanplaythrough = loaded;
+            audio.onerror = function() {
+                delete this.soundCache[file];
+                loaded();
+            }.bind(this);
+            audio.src = this.output.soundsDir + file + ".mp3";
+            
+            this.soundCache[file] = {
+                audio: audio,
+                __id: function () {
+                    return "getSound('" + file + "')";
+                }
+            };
+        }.bind(this));
     },
-
-
-    soundCache: {},
 
     // New methods and properties to add to the Processing instance
     processing: {
@@ -1656,16 +1696,10 @@ window.PJSOutput = Backbone.View.extend({
             var sound = this.soundCache[filename];
 
             if (!sound) {
-                var audio = document.createElement("audio");
-                audio.preload = "auto";
-                audio.src = this.output.soundsDir + filename + ".mp3";
-                sound = {audio: audio};
-                this.soundCache[filename] = sound;
+                throw {message:
+                    $._("Sound '%(file)s' was not found.", {file: filename})};
             }
-
-            sound.__id = function() {
-                return "getSound('" + filename + "')";
-            };
+            
             return sound;
         },
 
@@ -2011,9 +2045,8 @@ window.PJSOutput = Backbone.View.extend({
             }.bind(this));
         }.bind(this);
 
-        if (this.globals.getImage) {
-            this.cacheImages(userCode, runCode);
-
+        if (this.globals.getImage || this.globals.getSound) {
+            this.cacheResources(userCode, runCode);
         } else {
             runCode();
         }
--- a/build/js/live-editor.output_pjs.js
+++ b/build/js/live-editor.output_pjs.js
@@ -1152,126 +1152,156 @@ var BabyHint = {
 // TODO(jlfwong): Stop globalizing BabyHint
 window.BabyHint = BabyHint;
 
-var PJSResourceCache = Backbone.Model.extend({
+function PJSResourceCache(options) {
+    this.canvas = options.canvas;   // customized Processing instance
+    this.output = options.output;   // LiveEditorOutput instance
+    this.cache = {};
+    this.imageHolder = null;
+}
 
-    initialize: function(options) {
-        this.canvas = options.canvas;   // customized Processing instance
-        this.output = options.output;   // LiveEditorOutput instance
-        this.cache = {};
-        this.imageHolder = null;
-    },
+// Load and cache all resources (images and sounds) that could be used in 
+// the environment.  Right now all resources are loaded as we don't have 
+// more details on exactly which images will be required.
+//
+// Execution is delayed once a getImage/getSound appears in the source code
+// and none of the resources are cached. Execution begins once all the
+// resources have loaded.
+PJSResourceCache.prototype.cacheResources = function(userCode, callback) {
+    var resourceRecords = this.getResourceRecords(userCode);
 
-    // Load and cache all resources (images and sounds) that could be used in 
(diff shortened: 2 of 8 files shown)


build/js/live-editor.output_pjs.js
```diff
@@ -1577,12 +1585,16 @@ window.PJSOutput = Backbone.View.extend({
...
                 return loaded();
             }
-
+            
             file = fileMatch[1];
-
+            
+            if (this.imageCache[file]) {
+                return loaded();
+            }
```
kevinb7: changed so that ```this.imageCache[file] = img``` and ```if (this.imageCache[file]) {``` use the same file

build/js/live-editor.output_pjs.js
```diff
@@ -1602,10 +1614,38 @@ window.PJSOutput = Backbone.View.extend({
...
+                delete this.soundCache[file];
+                loaded();
+            }.bind(this);
+            audio.src = this.output.soundsDir + file + ".mp3";
+            
+            this.soundCache[file] = {
+                audio: audio,
+                __id: function () {
+                    return "getSound('" + file + "')";
+                }
```
kevinb7: only define the ```__id``` once for each sound

build/js/live-editor.output_pjs.js
```diff
@@ -1656,16 +1696,10 @@ window.PJSOutput = Backbone.View.extend({
             var sound = this.soundCache[filename];
 
             if (!sound) {
-                var audio = document.createElement("audio");
-                audio.preload = "auto";
-                audio.src = this.output.soundsDir + filename + ".mp3";
-                sound = {audio: audio};
-                this.soundCache[filename] = sound;
+                throw {message:
+                    $._("Sound '%(file)s' was not found.", {file: filename})};
```
kevinb7: throw if we can't find the sound

build/js/live-editor.output_pjs.js
```diff
@@ -1559,13 +1559,21 @@ window.PJSOutput = Backbone.View.extend({
...
+        }
 
         // Keep track of when image files are loaded
         var loaded = function() {
             numLoaded += 1;
 
             // All the images have loaded so now execution can begin
-            if (numLoaded === images.length) {
+            if (numLoaded === resourceCount) {
                 callback();
```
kevinb7: I'd like to rewrite the use promises at some point

build/js/live-editor.output_pjs.js
```diff
@@ -1540,9 +1539,10 @@ window.PJSOutput = Backbone.View.extend({
     // Execution is delayed once a getImage appears in the source code
     // and none of the images are cached. Execution begins once all the
     // images have loaded.
```
pamelafox: Can you update this comment as well?

build/js/live-editor.output_pjs.js
```diff
@@ -1577,12 +1585,16 @@ window.PJSOutput = Backbone.View.extend({
 
             // Skip if the image has already been cached
```
pamelafox: I think this comment needs to move down now?

build/js/live-editor.output_pjs.js
```diff
@@ -1602,10 +1614,38 @@ window.PJSOutput = Backbone.View.extend({
             // but PImage may be mutable, so that might not work.
             this.imageCache[file] = img;
         }.bind(this));
-    },
+        
+        _.each(sounds, function(file) {
```
pamelafox: Hm, interesting. I thought about doing this but worried about a program with a lot of sounds suffering from a long initial load. But I suppose its better for it to suffer in the beginning than later. We might want to have a more helpful loading indicator in future to specify that it's loading images and sounds, I believe SCRATCH does that, since their programs are so media-rich.

build/js/live-editor.output_pjs.js
```diff
@@ -1602,10 +1614,38 @@ window.PJSOutput = Backbone.View.extend({
...
+            var fileMatch = /["']([A-Za-z0-9_\/-]*?)["']/.exec(file);
 
+            if (!fileMatch) {
+                return loaded();
+            }
 
-    soundCache: {},
+            file = fileMatch[1];
+            
+            if (this.soundCache[file]) {
```
pamelafox: Observation: this logic seems very similar to image logic, it could be moved into a helper func?

build/js/live-editor.output_pjs.js
kevinb7: Totally agree.  Are there other places in KA that have a "loading resources..." progress bar?

pamelafox: We may have one other visual type of loader, the leaf, but the throbber is the standard visual. I think we could just put text underneath the throbber it to say what state its on ("Loading code, checking code, loading 3 images, loading 4 sounds, etc"). But that doesn't need to be done here.

build/js/live-editor.output_pjs.js
kevinb7: I have a couple of ideas, but I'd like to think through the pros/cons of each before proceeding.

kevinb7: I'd like to get your thoughts on this before proceeding:
- make the cached images look more like cached sounds, i.e. an object that looks like { image: imgElem, __id: function() { ... } }
- store all of the items in the same ```resourceCache``` where the object keys are the filenames including the extension (just incase there are some sounds with the same name as images), this will make it easy to have a single look up function
- use promises, this removes the need to count the number of resources we're going to load because we can wait for an array of promises to complete... not having to count means we can also remove the ```.match``` and just do the ```.exec```

kevinb7: I thought about it a little more and I'll do one loop to get all references and then a second loop to do initial all of the ```GET``` request.  That way we can display a message like 3 of 4 resources loaded.

pamelafox: That sounds good. Are you using in-browser promises or jquery deferred's or...?

kevinb7: I'm going to use jquery's deferred... sorry for not being specific.

build/js/live-editor.output.js
```diff
@@ -529,11 +529,13 @@ window.LiveEditorOutput = Backbone.View.extend({
...
-                    this.results.tests = testResults;
-                    this.phoneHome();
-                }.bind(this));
+                if (this.validate !== "") {
+                    this.test(userCode, this.validate, errors, function(errors, testResults) {
+                        this.results.errors = errors;
+                        this.results.tests = testResults;
+                        this.phoneHome();
+                    }.bind(this));   
+                }
```
kevinb7: This isn't necessary for resource caching.  It should've been included in the scrubber performance pull request.  I'm including it here because I don't want to forget about.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
 // TODO(jlfwong): Stop globalizing BabyHint
 window.BabyHint = BabyHint;
 
+var PJSResourceCache = Backbone.Model.extend({
```
kevinb7: Created a new module for all of the resource caching code.

build/js/live-editor.output_pjs.js
```diff
@@ -1221,6 +1342,11 @@ window.PJSOutput = Backbone.View.extend({
         this.bind();
  
         this.build(this.$canvas[0]);
+        
+        this.resourceCache = new PJSResourceCache({
+            canvas: this.canvas,
+            output: this.output
```
kevinb7: The reason why I'm passing the whole ```output``` object instead of just ```imagesDir``` and ```soundsDir``` is because ```setPaths``` is called asynchronously on the first run so we don't actually know the value for those paths yet.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
...
+    // more details on exactly which images will be required.
+    //
+    // Execution is delayed once a getImage/getSound appears in the source code
+    // and none of the resources are cached. Execution begins once all the
+    // resources have loaded.
+    cacheResources: function(userCode, callback) {
+        var getImageRegex = /getImage\s*\(['"](.*?)['"]\)/g;
+        var images = this.getRegexMatches(userCode, getImageRegex, 1);
+        var getSoundRegex = /getSound\s*\(['"](.*?)['"]\)/g;
+        var sounds = this.getRegexMatches(userCode, getSoundRegex, 1);
```
kevinb7: Originally I had separate methods called ```getSoundFilenames``` and ```getImageFilenames```... they we quite similar to so I decided to DRY the code out and ended up with this ```getRegexMatches``` method which I'm not sure is any better.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
...
+                    overflow: "hidden",
+                    position: "absolute"
+                })
+                .appendTo("body");
+        }
+
+        var promises = images.map(this.loadImage.bind(this));
+        promises = promises.concat(sounds.map(this.loadSound.bind(this)));
+
+        $.when.apply($, promises).then(callback);
```
kevinb7: This replaces the ```loaded()``` function.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
...
+                    height: 0,
+                    width: 0,
+                    overflow: "hidden",
+                    position: "absolute"
+                })
+                .appendTo("body");
+        }
+
+        var promises = images.map(this.loadImage.bind(this));
+        promises = promises.concat(sounds.map(this.loadSound.bind(this)));
```
kevinb7: I was thinking of having a single array of all resources to load with each item being an object with the filename and filetype and then this would turn into a single array mapped against ```loadResource``` which would multiplex between ```loadImage``` and ```loadSound```.  I'm going to give that a try to see that's any better.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
...
+        var deferred = $.Deferred();
+        var path = this.output.imagesDir + file + ".png";
+
+        var img = document.createElement("img");
+        img.onload = function() {
+            var pImg = new this.canvas.PImage(img);
+            pImg.__id = function() {
+                return "getImage('" + file + "')";
+            };
+            this.cache[file + ".png"] = pImg;
```
kevinb7: add the ```__id``` when here instead of in ```getImage```... all the tests pass so I think this should be fine.  We'll have to rethink this if we allow users to do pixel manipulation on images.  I'll add a comment to this effect.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,127 @@ var BabyHint = {
...
+    getSound: function(filename) {
+        var sound = this.cache[filename + ".mp3"];
+
+        if (!sound) {
+            throw {message:
+                $._("Sound '%(file)s' was not found.", {file: filename})};
+        }
+
+        return sound;
+    }
```
kevinb7: This could be combined into ```getResource(file,type)```.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,157 @@ var BabyHint = {
...
+};
+
+PJSResourceCache.prototype.loadImage = function(filename) {
+    var deferred = $.Deferred();
+    var path = this.output.imagesDir + filename + ".png";
+
+    var img = document.createElement("img");
+    img.onload = function() {
+        this.cache[filename + ".png"] = img;
```
kevinb7: reverted to caching  elements

kevinb7:
This should be good to go.

build/js/live-editor.output_pjs.js
```diff
@@ -1152,6 +1152,157 @@ var BabyHint = {
...
+PJSResourceCache.prototype.getResource = function(filename, type) {
+    switch (type) {
+        case "image":
+            return this.getImage(filename);
+            break;
+        case "sound":
+            return this.getSound(filename);
+            break;
+        default:
+            throw "we can't load '" + type + "' resources yet";
```
pamelafox: Please wrap in $._() for internationalization.

build/js/live-editor.output_pjs.js
pamelafox: Can you add that as a comment in the code, for future reference?

build/js/live-editor.output_pjs.js
kevinb7: A user should never see this.  All of the resources that are matched are either "sounds" or "images"... I was thinking of just doing ```break;``` for the default case.  Is there a way to log errors that users shouldn't see but might be useful for us back to a server?

build/js/live-editor.output_pjs.js
kevinb7: Will do.

build/js/live-editor.output_pjs.js
pamelafox: Ah, okay. We do automatically send JS errors off to Sentry though we dont necessarily use that information yet.

kevinb7: I had a quick search through live-editor and it doesn't look like we're using Sentry anywhere.  I'm going to get rid of the throw and put a TODO there instead and track it with a Github issue for add error tracking via Sentry.

pamelafox: Oh, I think we're likely hooking in globally on KA through window.onerror or such. Anyway, a break is fine. I'll find out about issues from students and ZenDesk tickets.

kevinb7: ```break``` it is.

pamelafox:
Is this PR ready to go? I can merge/release it this week if so.

kevinb7:
I believe this is good to go.

PR #245, opened 2015-02-07
Json worker deps

kevinbarabash:


build-paths.json
```diff
@@ -71,7 +71,8 @@
             "external/jshint/jshint.js",
             "external/structuredjs/external/esprima.js",
             "external/structuredjs/structured.js",
-            "external/TraceKit/tracekit.js"
+            "external/TraceKit/tracekit.js",
+            "external/blob/blob.js"
```
kevinb7: polyfill Blobs on IE9

build/js/live-editor.output.js
```diff
@@ -6,18 +6,13 @@ this["Handlebars"]["templates"]["output"] = Handlebars.template(function (Handle
...
     this.pool = [];
     this.curID = 0;
-    this.filename = filename;
+    this.url = url;
     this.onExec = onExec || function() {};
 };
 
-PooledWorker.prototype.getURL = function() {
-    return this.workersDir + this.filename +
-        "?cachebust=G" + (new Date()).toDateString();
```
kevinb7: Any cache busting will need to be done directly on the request for ```deps.json```.

build/js/live-editor.output_pjs.js
```diff
@@ -1213,15 +1213,27 @@ window.PJSOutput = Backbone.View.extend({
         this.config = options.config;
         this.output = options.output;
 
-        this.tester = new PJSTester(_.extend(options, {
-            workerFile: "pjs/test-worker.js"
-        }));
```
kevinb7: moved to ```initializeWorkers```

build/js/live-editor.output_pjs.js
```diff
@@ -1213,15 +1213,27 @@ window.PJSOutput = Backbone.View.extend({
...
+        location.search.substring(1).split("&").forEach(function(param) {
+            var parts = param.split("=");
+            params[parts[0]] = decodeURIComponent(parts[1]);
+        });
+
+        // Initialize workers:
+        // The reason why workersDir is passed as a URL parameter is so 
+        // that pjs-output.js can do a GET request for the worker code and
+        // their dependencies without having to wait for the initial message
+        // from the parent frame that contains all of the paths.
```
kevinb7: In the original version of this feature I injected the path directly into output.html.  I don't know what the path is on the server and so I'm using the ```workersDir``` path the live-editor.js has.  This was the most direct approach I could think of.

js/workers/pjs/jshint-worker.js
```diff
@@ -1,24 +1,27 @@
...
-        importScripts(event.data.jshintFile +
-            "?cachebust=" + (new Date()).toDateString());
+            eval(deps["es5-shim.js"]);
+            eval(deps["jshint.js"]);
+            eval(deps["underscore.js"]);
 
-        importScripts(event.data.externalsDir +
-            "underscore/underscore.js?cachebust=" + (new Date()).toDateString());
+            self.JSHINT = JSHINT;    // store it in a global so we can use it next time
+        }
```
kevinb7: ```throw``` if we don't get the dependencies?

tests/output/pjs/index.html
```diff
@@ -6,6 +6,24 @@
...
+            // their dependencies without having to wait for the initial message
+            // from the parent frame that contains all of the paths.
+            if (location.search.indexOf("workersDir=") === -1) {
+                var workersDir = encodeURIComponent("../../../build/workers/");
+                if (location.search.indexOf("?") === -1) {
+                    location.search = "?workersDir=" + workersDir;
+                } else {
+                    location.search += "&workersDir=" + workersDir;
+                }
+            }
```
kevinb7: The tests don't load ```output_pjs.js``` using an ```iframe``` so we have to add a URL parameter ourselves.

kevinb7:
The code in ```deps.json``` has not been minified.  We probably need to add a step for that in our build process because chances are the step that minifies .js files later in the build pipeline will ignore the .json files.

bundle_deps.js
```diff
@@ -0,0 +1,67 @@
...
+        // tasks concurrently is the order that they complete in is random.
+        // The interesting thing about the order of keys in a dictionary
+        // is that it's not alphabetical, but rather based on order of addition
+        // to the dictionary. 
+        return q.all(
+            Object.keys(deps).map(function (filename) {
+                var path = deps[filename];
+                return readFile(path, { encoding: 'utf8' }).then(function (contents) {
+                    var result = uglify.minify(contents, {fromString: true});
+                    obj[filename] = result.code;
```
kevinb7: Uglify each file in the bundle.  This reduced the size of the bundle from 635KB to 239KB which is awesome because neither the workers nor their dependencies are currently being minified in production.  We can get this number even lower because none of the global symbols (or properties names on global objects) aren't being mangled, but it's going to take additional work that I'd like to defer for now.

build/js/live-editor.output.js
```diff
@@ -529,11 +533,13 @@ window.LiveEditorOutput = Backbone.View.extend({
...
-                    this.results.errors = errors;
-                    this.results.tests = testResults;
-                    this.phoneHome();
-                }.bind(this));
+                if (this.validate !== "") {
+                    this.test(userCode, this.validate, errors, function(errors, testResults) {
+                        this.results.errors = errors;
+                        this.results.tests = testResults;
+                        this.phoneHome();
+                    }.bind(this));
```
kevinb7: Only test the code using the test worker if ```validate``` is not empty.  This saves cycles when the user is not doing a challenge.

kevinb7:
screen shot 2015-07-06 at 10 01 13 am
It took 13s to load for a cold load.  The green bars are waiting.  We could definitely reduce this quite a bit by packaging stuff together as per this pull request.

Issue #256, opened 2015-02-18
Webpages not updating immediately after changing the color


pamelafox:
Open https://www.khanacademy.org/computer-programming/results-for-contest-fabric-pattern/6250308077092864 in Chrome, try changing rgb(), color does not update immediately.

Seems like a regression bug, possibly due to recent number scrubber change from @kevinb7. Haven't looked into it yet.

MikaalSky:
The page is updating, though. Notice the flashing of link colors.

MikaalSky:
Alright, this is happening in Firefox, also. An error is getting fired.
TypeError: list is undefined
If you can't replicate, move the color selector circle around very fast.

kevinb7:
I will look into this.

PR #258, opened 2015-02-18
fix scrubbing in non pjs environments

kevinb7:
Don't skip linting in non pjs environments.

### DIFF
--- a/build/js/live-editor.ui.js
+++ b/build/js/live-editor.ui.js
@@ -928,14 +928,16 @@ window.LiveEditor = Backbone.View.extend({
             type: this.editorType
         });
         
-        this.noLint = false;
-        this.editor.on("scrubbingStarted", function() {
-            this.noLint = true;
-        }.bind(this));
-        
-        this.editor.on("scrubbingEnded", function() {
+        if (this.editorType === "ace_pjs") {
             this.noLint = false;
-        }.bind(this));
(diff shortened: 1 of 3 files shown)


kevinb7:
linting in the webpage environment generates slowparseResults which are then used in the runCode step so skipping linting won't work in that environment without some more work.

kevinb7:
Addresses issue https://github.com/Khan/live-editor/issues/256.

pamelafox:
We're not particularly well set up to be able to test this, ay? 
Can you add this comment to the code above the if?
"linting in the webpage environment generates slowparseResults which are then used in the runCode step so skipping linting won't work in that environment without some more work."

kevinb7:
With some work we could test things like that.  I've done lots of stuff with simulating mouse events and checking pixels and the DOM.  As a start though I'd like to automate the tests that we have using travis-ci.  Once that's in place I think it makes more sense to tackle writing some functional tests.

kevinb7:
@pamelafox is there some part of the pipeline that is already running the existing tests automatically?

pamelafox:
Good Q, answer is no, not currently. Travis-CI would make sense for this repo, I think it's free for OSS projects? 
Internally we use mocha/chai/sinon plus phantomJS, with jenkins automation.

PR #259, opened 2015-02-18
automate tests with travis-ci

kevinb7:
I was trying to get this working so that we could run the tests in Chrome on travis-ci, but for some reason the test runner won't start Chrome in that environment (works fine on OS X).  When I have more time I'll set up an ubuntu virtual machine and see if I can figure out what's going on.  In the meantime though, I've automated the tests that I can using phantomjs.  You'll have to enable access on travis-ci to khan/live-editor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ before_script:
   - npm install -g bower
   - bower install
 script:
+  - npm run build
   - ./node_modules/.bin/testee tests/output/pjs/index.html
   - ./node_modules/.bin/testee tests/output/webpage/index.html
   - ./node_modules/.bin/testee tests/tooltips/index.html
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+node_js:
+  - "0.10"
+before_script:
+  - npm install -g bower
+  - bower install
+script:
+  - ./node_modules/.bin/testee tests/output/pjs/index.html
+  - ./node_modules/.bin/testee tests/output/webpage/index.html
+  - ./node_modules/.bin/testee tests/tooltips/index.html
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "Live Code Editor",
   "main": "dist/live-editor.js",
   "scripts": {
+    "build": "gulp build",
     "test": "gulp test"
   },
   "repository": {
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "async": "^0.9.0",
     "request": "^2.44.0",
     "sinon": "^1.10.3",
+    "testee": "^0.1.8",
     "handlebars": "1.0.5-beta",
     "gulp-bower": "0.0.7",
     "gulp-symlink": "^2.0.1"
--- a/tests/output/pjs/output_test.js
+++ b/tests/output/pjs/output_test.js
@@ -643,6 +643,9 @@ describe("Scratchpad Output Exec", function() {
             Program.restart();
         },
         errors: [],
+        // p.Program methods weren't being stubbed and were causing some of the
+        // test code to be run in an infinite loop after the it() block had 
+        // completed.
         setup: function(output) {
             var p = output.output.canvas;
             sinon.stub(p.Program, "settings");
--- a/tests/output/pjs/output_test.js
+++ b/tests/output/pjs/output_test.js
@@ -644,7 +644,7 @@ describe("Scratchpad Output Exec", function() {
         },
         errors: [],
         // p.Program methods weren't being stubbed and were causing some of the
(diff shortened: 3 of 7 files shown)


tests/output/pjs/output_test.js
```diff
@@ -637,11 +637,24 @@ describe("Scratchpad Output Exec", function() {
...
             Program.restart();
             Program.runTests(function() {});
         },
-        errors: []
+        errors: [],
+        setup: function(output) {
+            var p = output.output.canvas;
+            sinon.stub(p.Program, "settings");
+            sinon.stub(p.Program, "restart");
+            sinon.stub(p.Program, "runTests");
```
kevinb7: processing.Program methods weren't being stubbed and where actually causing certain test code to be run in an infinite loop after ```it``` had completed.

kevinb7: I will add this comment to output_test.js

tests/output/webpage/output_test.js
```diff
@@ -149,18 +149,6 @@ describe("Linting", function() {
...
-            {row: 0, column: 0, lint: {type: "ELEMENT_NOT_ALLOWED"}}
-        ]
-    );
-
-    failingTest("Video element banned",
-        "<video></video>", [
-            {row: 0, column: 0, lint: {type: "ELEMENT_NOT_ALLOWED"}}
-        ]
-    );
-
```
kevinb7: These tests were failing because "audio" and "video" are in the ```disableTags``` array in webpage-output.js.  I figured that this was intentional, that's why I removed these tests.

tests/tooltips/index.html
```diff
@@ -43,11 +43,13 @@
...
-            } else {
-                mocha.run();
-            }
+            onload = function(){
+                if (window.mochaPhantomJS) {
+                    mochaPhantomJS.run();
+                } else {
+                    mocha.run();
+                }
+            };
```
kevinb7: testee.js require an ```onload``` function like this to run correctly.

package.json
```diff
@@ -33,6 +33,7 @@
     "async": "^0.9.0",
     "request": "^2.44.0",
     "sinon": "^1.10.3",
+    "testee": "^0.1.8",
```
kevinb7: I've had much better luck with this test runner than with karma.

tests/output/pjs/output_test.js
```diff
@@ -637,11 +637,27 @@ describe("Scratchpad Output Exec", function() {
     runTest({
         title: "Program functions exist",
         code: function() {
-            var x = Program.settings();
+            debug("start");
```
pamelafox: What's the debug() for?

kevinb7: whoopsies... that was from when I trying to diagnose the infinite loop issue below.   I'll clean that up.

kevinb7:
@pamelafox this should be good to go

pamelafox:
Can you resolve the conflicts in this one? (Probably due to me removing the audio/video test since you sent it). Thanks!

kevinb7:
I will, sometime tomorrow.

kevinb7:
@pamelafox sorry, it'll have to wait for the weekend.

pamelafox:
No worries!

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+node_js:
+  - "0.10"
+before_script:
+  - npm install -g bower
+  - bower install
+script:
+  - ./node_modules/.bin/testee tests/output/pjs/index.html
+  - ./node_modules/.bin/testee tests/output/webpage/index.html
+  - ./node_modules/.bin/testee tests/tooltips/index.html
```
kevinb7: right now testee offers no benefit over 'npm run test' b/c it uses phantomjs by default.  I've had good success running firefox via testee on travis-ci, but unfortunately, our test suites do not pass in firefox.  I tried getting this to work with chrome w/o success.  setting up a unix environment to debug what's wrong with the chrome launcher is on my list of things to do.  It would also be good to fix the tests so that they work in firefox too... being able to run the tests automatically in both browsers would be ideal.

.travis.yml
```diff
@@ -0,0 +1,10 @@
+language: node_js
+node_js:
+  - "0.10"
+before_script:
+  - npm install -g bower
+  - bower install
```
kevinb7: this isn't really necessary b/c the tests are run against built code and we're not building the code though we really should.

kevinb7:
I'd like to update this so that we're also building each time we run the tests before merging.

.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+script:
+  - npm run build
```
kevinb7: build before running the tests

PR #315, opened 2015-06-10
Added Aqualine to the Image Picker

JavascriptFTW:
I added Aqualine to the image picker since all the other avatars that the user can user are already there so I didn't see why Aqualine shouldn't be as well.
I tried to follow the pattern (is there a pattern?) of having two "regular" avatars between each swarm of seed-seedling-sapling-tree-ultimate-ness.
If there is a specific reason that Aqualine shouldn't be in the image picker I understand, but if there is please let me know what that reason is so that I don't make a pull request like this in the future


pamelafox:
Can you also add the Aqualine images to the images dir?

JavascriptFTW:
I looked in the live-editor\images\avatars dir and Aqualine was already there _before_ I did this. (along with not yet released avatars like Starky and Blue Spice Man). Perhaps I don't understand what you mean? Is there some other place I need to add Aqualine to?

JavascriptFTW:
I added Aqualine to the build folder. I think I've added her(?) everywhere now. (js, build, images)

kevinb7:
@JavascriptFTW Cool!  One last thing, you'll need to sign a CLA.  http://www.khanacademy.org/r/cla

JavascriptFTW:
Two questions before I agree to that. One:I am not a legal adult so should I get a parent/guardian to agree to it for me? Two:Is the "Electronic Signature" supposed to be some username that I use online?

kevinb7:
I don't know the answer to question one, but I will find out.  As for the second question, you just have to type "I AGREE" in the box to signify that you agree.

JavascriptFTW:
I'll wait to agree until you answer the first question, I don't want to mess something up legally for you guys.

kevinb7:
@JavascriptFTW Thank you for your patience while we figure this out.

kevinb7:
@JavascriptFTW Please have a parent/guardian fill out the form and enter your username in username field.

JavascriptFTW:
My Mom filled out the form, and we weren't sure whose name to put in the name box so she put mine and then a note in the mailing address. She said that if you need verification to email her (will supply email address only if needed).

kevinb7:
I think that's good.  Thanks for filling out the form now LGTM!  ^_^

kevinb7:
Works like a charm!  I'm sure all the Aqualine enthusiasts will be quite happy with this addition to the image picker.

screen shot 2015-06-10 at 6 32 04 pm

kevinb7:
@JavascriptFTW FYI, although your changes have been merged into live-editor, they won't show up on khanacademy.org until we pull the changes into our web app project and deploy it and I don't know when that will happen next.  I'll try to make it happen sooner rather than later.

JavascriptFTW:
I don't really care. I've got another pull request I want to make sometime this month anyway (won't be more avatars :P)!

JavascriptFTW:
I have a stupid question. How do I set it up so that I can test my local changes in the live editor to make sure that I haven't broken something before I go off submitting a buggy PR? I looked at the instructions in the README and I don't understand, do I need to set up a server for it to work? Whenever I try to run the HTML files in the live-editor dir I just get the throbber.gif going on forever. I'm probably missing something and the answer is in front of my face.

pamelafox:
It will need to run off localhost, so that its on http protocol.
Have you tried using python simple server? http://www.linuxjournal.com/content/tech-tip-really-simple-http-server-python

JavascriptFTW:
Thanks. I got it running now. I really need to get some more practice with this kind of stuff.

PR #347, opened 2015-07-06
Avoid undo avalanche on number-scrubbing

ChrisJPhoenix:
First-draft code. 
Test doesn't work; see comments. 
Ugliness if user hits a key while scrubbing; see comments. 
Not implemented for color picker yet.

### DIFF
--- a/check.js
+++ b/check.js
@@ -1,0 +1,50 @@
+// Exports a function to check that all source files and dependencies exist.
+
+var fs = require("fs");
+
+module.exports = function() {
+    var missing = [];
+    var traverse = function(obj) {
+        if (Array.isArray(obj)) {
+            obj.forEach(function (path) {
+                if (path.indexOf("*") === -1) {
+                    try {
+                        fs.statSync(path);
+                    } catch(e) {
+                        missing.push(path);
+                    }
+                }
+            });
+        } else {
+            for (var i in obj) {
+                if (obj.hasOwnProperty(i)) {
+                    traverse(obj[i]);
+                }
+            }
+        }
+    };
+
+    traverse(require("./build-paths.json"));
+
+    if (missing.length > 0) {
+        var bower = false;
+        var submodule = false;
+        console.log("missing files:");
+        missing.forEach(function (path) {
+            console.log("  " + path);
+            if (path.indexOf("bower_components") !== -1) {
+                bower = true;
+            } else {
+                submodule = true;
+            }
+        });
+        console.log("make sure to run:");
+        if (bower) {
+            console.log("  bower install");
+        }
+        if (submodule) {
+            console.log("  git submodule sync && git submodule update");
+        }
+    }
+    return missing;
+};
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -16,6 +16,7 @@ var mochaPhantomJS = require("gulp-mocha-phantomjs");
 var staticServe = require("node-static");
 var request = require("request");
 
+var check = require("./check.js");
 var paths = require("./build-paths.json");
 
 gulp.task("templates", function() {
@@ -197,7 +198,18 @@ gulp.task("test", function(callback) {
     runSequence("test_output_pjs", "test_output_webpage", "test_tooltips", callback);
 });
 
+// Check to make sure all source files and dependencies exist before building.
+gulp.task("check", function() {
+    var missing = check();
+    if (missing.length > 0) {
+        console.log("Aborting build");
+        process.exit();
+    } else {
+        console.log("all files exist");
+    }
+});
+
 gulp.task("build",
-    ["templates", "scripts", "workers", "styles", "images", "externals"]);
+    ["check", "templates", "scripts", "workers", "styles", "images", "externals"]);
 
 gulp.task("default", ["watch", "build"]);


js/ui/tooltips/number-scrubber.js
```diff
@@ -79,6 +79,14 @@ TooltipEngine.classes.numberScrubber = TooltipBase.extend({
...
                     $(this).css("visibility", "hidden");
                     self.trigger("scrubbingStarted");
+                    // The text-to-be-tweaked needs to be the same length at the start and end
+                    // of the anti-undo changes.
+                    // I could probably just remember the length, but I like putting back the
+                    // original string. (It might even matter for i18n.)
+                    var Range = ace.require("ace/range").Range;
+                    var loc = self.aceLocation;
+                    var range = new Range(loc.row, loc.start, loc.row, loc.start + loc.length);
+                    self.originalString = parent.editor.getTextRange(range);
```
kevinb7: This line doesn't work for me.  You're using the global ```parent``` here which is defined to be ```window``` for some reason.  Also, there is no ```getTextRange``` on editor.  Try:
```
self.originalString = self.parent.editor.getSession().getTextRange(range);
```

tests/tooltips/tooltips_test.js
```diff
@@ -22,4 +22,32 @@ describe("General Tooltip Tests", function(){
...
+        var cursor = editor.getCursorPosition();
+        var Range = ace.require("ace/range").Range;
+        var range = new Range(cursor.row, cursor.column-3, cursor.row, cursor.column-2);
+        editor.session.replace(range, "6");
+        expect(getLine()).to.be.equal("a(6);");
+        editor.session.replace(range, "7");
+        expect(getLine()).to.be.equal("a(7);");
+        editor.undo();
+        // This next test will fail because undo() clears the buffer, undoing all changes.
+        // I need a way to keep ACE from merging the deltas.
```
kevinb7: Try calling the same method the number scrubber call, i.e. ```TTE.updateText```.

tests/tooltips/tooltips_test.js
```diff
@@ -22,4 +22,32 @@ describe("General Tooltip Tests", function(){
         TTE.remove();
         expect($('.tooltip').length).to.be.equal(0);
     });
-});
\ No newline at end of file
+
+    it("ACE anti-undo hack still works", function() {
```
kevinb7: These tests should go in numberScrubber_test.js.

kevinb7:
> Ugliness if user hits a key while scrubbing; see comments. 

We should either:
- ignore keyboard input while scrubbing
- cancel scrubbing on keyboard input

I think the first one is probably a better experience b/c if you hit the keyboard while scrubbing it was probably an accident.

> Not implemented for color picker yet.

This would be awesome!

In playing around with it I noticed that when you undo after scrubbing it hides the scrubber and shows a tooltip with the names of the parameters.  We should show the scrubber after an undo in this situation b/c that's what the user saw before they started scrubbing and ```undo()``` is just returning the user to that state.

js/ui/tooltips/number-scrubber.js
ChrisJPhoenix: Agh, sorry. I couldn't get gulp to work (I don't have jQuery files installed, etc. - I found the link to onboarding install info but didn't have permission to access it) so I was hacking up build/js/live-editor.tooltips.js and then trying (imperfectly) to keep it in sync with number-scrubber.js etc. So I did test it, and it worked, and then I didn't copy it back. (I had editor.session rather than editor.getSession() - I'll fix that.)

kevinb7: Did you run the commands in the https://github.com/khan/live-editor#building?  If so, did you get any errors when trying to run them?

tests/tooltips/tooltips_test.js
ChrisJPhoenix: Eventually the color picker will use this hack too...? That's why I put it here instead of there.

kevinb7: That makes sense.  Disregard my previous comment.

ChrisJPhoenix:
I realized over dinner that I could cancel scrubbing on keyboard input. To ignore keyboard input while scrubbing - this thread https://groups.google.com/forum/#!topic/ace-discuss/AoKWYE2vhbg implies that I can do editor.keyBinding.setDefaultHandler(null) but google for 
   keybinding site:ace.c9.io 
finds nothing, so I'm a bit unsure about using it.

kevinb7:
You could use ```editor.setReadOnly(true)``` to disable modifications from the keyboard while scrubbing.  I did a little testing and it doesn't affect ```session.replace()``` or ```undo()``` for that matter.  You should probably also check what happens when you active ```undo``` while scrubbing.  It should probably put the number back to what it was at the start.

js/ui/tooltips/number-scrubber.js
ChrisJPhoenix: TBH I'd been trying to resolve the errors as they came up - I tried those commands and now I'm getting a different error. I ran the suggested git commands and still got the same error. (I'm running on Windows.)

F:\khan\live-editor-fork>gulp
[10:24:13] Using gulpfile F:\khan\live-editor-fork\gulpfile.js
[10:24:13] Starting 'watch'...
[10:24:13] Finished 'watch' after 628 ms
[10:24:13] Starting 'check'...
missing files:
  external/iframe-overlay/dist/iframe-overlay.js
  external/iframe-overlay/dist/iframe-overlay.js
make sure to run:
  git submodule sync && git submodule update
Aborting build

F:\khan\live-editor-fork>

kevinb7: I need to fix that suggestion message.  I forgot that not everything in external/ comes from submodules.  That particular dependency comes from bower.  If you don't have bower installed you can install it by running: ```npm install -g bower```.  Then run ```bower install``` to install the dependencies.

ChrisJPhoenix:
Whew... the last 10% of the move took the other 90% of the time, and now I'm backlogged on other stuff. I will finish this - just not this week.

kevinb7:
@ChrisJPhoenix no worries.  It's always that last 10%.  Thanks for the update!

ChrisJPhoenix:
I'm getting phase-of-the-moon flakiness when testing undo. Here's the code that is flaky:

describe("tooltips_test - test ACE anti-undo hack", function() {
    it("ACE anti-undo hack still works", function() {
        TTE.editor.setValue("");
        TTE.editor.onTextInput("a(12");
        expect(TTE.currentTooltip).to.be.equal(TTE.tooltips.numberScrubber);
        TTE.currentTooltip.updateText("14");
        // Should be a(14
        expect(getLine()).to.be.equal("a(14");
        TTE.currentTooltip.updateText("24");
        // should be a(24
        expect(getLine()).to.be.equal("a(24");
        TTE.editor.undo();
        expect(getLine()).to.be.equal("a(24"); // Huh? It didn't undo!
...

And here are some typical results (both displayed on the same test webpage in a single test run - I put the above code in two different files)

Error: expected '' to equal 'a(24'
    at Assertion.assert (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:96:13)
    at Assertion.be.Assertion.equal (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:216:10)
    at Function.scope (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:499:17)
    at Context.<anonymous> (http://localhost:63342/live-editor-fork/tests/tooltips/colorPicker_test.js:134:33)

Error: expected 'getImage("a);' to equal 'a(24'
    at Assertion.assert (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:96:13)
    at Assertion.be.Assertion.equal (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:216:10)
    at Function.scope (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:499:17)
    at Context.<anonymous> (http://localhost:63342/live-editor-fork/tests/tooltips/tooltips_test.js:13:33)

Sometimes the test passes that line - i.e. the "It didn't undo!" comment is correct - some of the time.

js/ui/tooltips/number-scrubber.js
ChrisJPhoenix: I had run 'bower install' - it was in the link you sent. When I run it again, it prints no output so I guess it's happy - and I still get the above error output from 'gulp'.

kevinb7:
You might have to wait a bit before actually checking the values.  Try something like this:
```javascript
it("ACE anti-undo hack still works", function() {
        TTE.editor.setValue("");
        TTE.editor.onTextInput("a(12");
        expect(TTE.currentTooltip).to.be.equal(TTE.tooltips.numberScrubber);
        TTE.currentTooltip.updateText("14");
        // Should be a(14
        expect(getLine()).to.be.equal("a(14");
        TTE.currentTooltip.updateText("24");
        // should be a(24
        expect(getLine()).to.be.equal("a(24");
        TTE.editor.undo();
        setTimeout(function () {
                expect(getLine()).to.be.equal("a(24"); // Huh? It didn't undo!
                done();
        }, 100);
});
```
Can you push your changes so I can try it out?

I formatted the code snippets in your comment.  You can do the same: https://help.github.com/articles/github-flavored-markdown/.

js/ui/tooltips/number-scrubber.js
kevinb7: Check that __bower_components/iframe-overlay/dist/iframe-overlay.js__ exists. My guess is that windows and/or git for windows is having troubles with the symlinks.  

You might try the first solution in http://stackoverflow.com/questions/5917249/git-symlinks-in-windows.  Alternately you could modify build-paths.json and replace __external/iframe-overlay/dist/iframe-overlay.js__ with __bower_components/iframe-overlay/dist/iframe-overlay.js__.  

I think the later is probably the better approach b/c it solves the problem for other windows users.

ChrisJPhoenix:
Thanks for the code-formatting pointer - I'll do that from now on.

I've pushed my changes. 

I don't understand your suggestion of using a delay to check the undo result: since TTE.editor is re-used between tests, I'd think the callback would get some completely random state after several more tests have run. Is there a way to tell the editor "Run synchronously all the queued changes"?

BTW, at some point I remember you suggested I use a different git workflow (fork from a different branch?), but it looks like you've pruned the comment threads because I don't see your suggestion anymore. Did you want me to change my source branch for this project, or just for future work on other features? (I'm very interested in adding a feature to #include one program in another, and also happy to work on features you suggest.)

kevinb7:
> Is there a way to tell the editor "Run synchronously all the queued changes"?

There is.  Change the callback to `it` to take a `done` parameter.  I update my previous comment b/c I forgot about it.  The mocha docs are pretty sparse, but they have some examples: http://mochajs.org/#asynchronous-code.

> Did you want me to change my source branch for this project, or just for future work on other features?

I find it makes things easier from a contributor's perspective.  It was more of a suggestion for the future pull requests.  I think you'd have to start a new pull request if you tried to do it for this feature.

kevinb7:
@ChrisJPhoenix I think I've fixed the issue with gulp complaining.  They're on master.  Let me know if you're still having problems with gulp afterwards.

I messed around with the tests last night.  Adding a timeout and running the new tests in isolation seems to produce reliable passes.  Running with the rest of tests produces failures with data bleeding through from other tests.  The problem is that we create one instance and re-use that for all of the tests.  This is bad practice.  The output tests create a new LiveEditorOutput instance for each test.  We should be doing the same thing for the tooltip tests.

ChrisJPhoenix:
I wrote a function which adds a unique editor to the page. The trouble is that when I call it, another test fails - the "General -> Remove succeeds" test:
```
Error: expected 5 to equal 0
    at Assertion.assert (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:96:13)
    at Assertion.be.Assertion.equal (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:216:10)
    at Function.<anonymous> (http://localhost:63342/live-editor-fork/node_modules/expect.js/index.js:499:17)
    at Context.<anonymous> (http://localhost:63342/live-editor-fork/tests/tooltips/tooltips_test.js:52:44)
```
If I comment out line 3 of tooltips_test.js then that test passes. I'm guessing there's some global state in the tooltips code?

Separately - the 'done' parameter/function seems to be only for the purpose of telling Mocha that your test has reached the end of its control flow. It doesn't seem to call async / event loop code - merely that it's supposed to be run after at the end of whatever async code you've invoked in your test. 

I'm thinking that a good example to copy is the stop() function in number-scrubber.js (around line 122): (And there I would want to use the 'done' parameter). Maybe this is what you were getting at?
```js
                    // use a timeout because $leftButton.click and $rightButton.click
                    // are called after stop
                    setTimeout(function () {
                        self.dragged = false;
                        self.parent.editor.setReadOnly(self.wasReadOnly);
                    }, 0);
```

So if I wanted to test several undo's in one function, I'd need to use nested callbacks...? It seems like updateText is synchronous but undo() is asynchronous...? That's weird. I feel like I'm missing something. 

I'll keep working on writing a good test, and putting the undo fix in the color picker, on the assumption that we can get the test framework working.

kevinb7:
@ChrisJPhoenix thanks for sticking with this, I think we're almost there.

> another test fails - the "General -> Remove succeeds" test
> I'm guessing there's some global state in the tooltips code?

Good guess.  The unique editor has add stuff to the DOM that needs to be cleaned up.  See my next post for how to do this.

> Separately - the 'done' parameter/function seems to be only for the purpose of telling Mocha that your test has reached the end of its control flow.

There are two parts to writing async tests: adding a `done` parameter and calling the `done` callback.  The first indicates to mocha that this is an async test and that it should pass a `done` function to our callback.  The second let's mocha know that the test has completed.  If we don't call `done` the test will timeout.

> It seems like updateText is synchronous but undo() is asynchronous...? That's weird. I feel like I'm missing something.

That's about right.  It turns out you have to wait before calling `undo()` otherwise the undo stack doesn't get updated.  I'm not sure why ace does this asynchronously.

kevinb7:
I was able to get the tests working.  I made some changes to the shared.js as well.  Could you work these changes into you pull request?

__tooltips_test.js__
```javascript
// complete file
describe("tooltips_test - test ACE anti-undo hack", function() {
    var editor, tte, session;
    
    beforeEach(function () {
        var ue = uniqueEditor(); 
        editor = ue.editor;
        tte = ue.tooltipEngine;
        session = editor.getSession();
    });
    
    afterEach(function (done) {
        setTimeout(function() {
            tte.remove();
            done();
        }, 0)
    });
    
    it("ACE anti-undo hack still works", function(done) {
        editor.setValue("");
        editor.onTextInput("a(12");
        expect(tte.currentTooltip).to.be.equal(tte.tooltips.numberScrubber);
        setTimeout(function () {
            tte.currentTooltip.updateText("14");
            expect(session.getValue()).to.be.equal("a(14");
            setTimeout(function () {
                tte.currentTooltip.updateText("24");
                expect(session.getValue()).to.be.equal("a(24");
                setTimeout(function() {
                    editor.undo();
                    expect(session.getValue()).to.be.equal("a(14");
                    setTimeout(function() {
                        editor.undo();
                        expect(session.getValue()).to.be.equal("a(12");
                        setTimeout(function() {
                            editor.undo();
                            expect(session.getValue()).to.be.equal("");
                            done();
                        }, 0);
                    }, 0);
                }, 0);
            }, 0);
        }, 0);
    });
    
    it("should undo multiple changes with avoidUndo = true", function (done) {
        editor.setValue("");
        editor.onTextInput("a(12");
        expect(tte.currentTooltip).to.be.equal(tte.tooltips.numberScrubber);
        setTimeout(function () {
            tte.currentTooltip.updateText("14");
            expect(session.getValue()).to.be.equal("a(14");
            setTimeout(function () {
                tte.currentTooltip.updateText("24", null, true);
                expect(session.getValue()).to.be.equal("a(24");
                setTimeout(function () {
                    tte.currentTooltip.updateText("34", null, true);
                    expect(session.getValue()).to.be.equal("a(34");
                    setTimeout(function() {
                        editor.undo();
                        expect(session.getValue()).to.be.equal("a(12");
                        setTimeout(function() {
                            editor.undo();
                            expect(session.getValue()).to.be.equal("");
                            done();
                        }, 0);
                    }, 0);
                }, 0);
            }, 0);
        }, 0);
    });
});

describe("General Tooltip Tests", function(){
    it("autoSuggest - detection", function() {
        typeLine("rect(44");
        expect(TTE.currentTooltip).to.be.equal(TTE.tooltips.autoSuggest);
    });

    it("autoSuggest vs numberScrubber (click)", function() {
        var line = "rect(12, 34, 56, 78);";
        var pre = "rect(12";
            
        var e = getTooltipRequestEvent(line, pre);
        TTE.editor._emit("requestTooltip", e);
        expect(TTE.currentTooltip).to.be.equal(TTE.tooltips.autoSuggest);

        e = getTooltipRequestEvent(line, pre); // Reset propagationStopped
        e.source = {action: "click"};
        TTE.editor._emit("requestTooltip", e);
        expect(TTE.currentTooltip).to.be.equal(TTE.tooltips.numberScrubber);
    });

    it("Remove succeeds", function(done) {
        TTE.remove();
        setTimeout(function () {
            expect($('.tooltip').length).to.be.equal(0);
            done();
        }, 0)
    });

});
```

__shared.js__ (I cleanup `uniqueEditor()` a bit and now use it to create `window.ACE`):
```javascript
// start of file
var mockObject = function(obj, mocks) {
    _.each(mocks, function(method) {
        obj[method] = sinon.spy();
    });
    return obj;
};

window.tooltipClasses = TooltipEngine.classes;

var uniqueEditor = function() {
    var elem = document.createElement('div');
    document.body.appendChild(elem);
    var ace = new AceEditor({ //Initializes TooltipEngine internally
        el: elem,
        autoFocus: true,
        config: new ScratchpadConfig({
            version: 3
        }),
        imagesDir: "../../build/images/",
        externalsDir: "",
        workersDir: "",
        record: new ScratchpadRecord(),
        type: "ace_pjs"
    });
    ScratchpadAutosuggest.init(ace.editor);
    ace.editor.focus();
    ace.setSelection({
        start: {
            row: 0,
            column: 0
        },
        end: {
            row: 0,
            column: 0
        }
    });
    return ace;
};

window.ACE = uniqueEditor();
window.editor = ACE.editor;
window.TTE = ACE.tooltipEngine;

for (var TooltipName in TooltipEngine.prototype.classes) {
    var Tooltip = TooltipEngine.classes[TooltipName];
    Tooltip.prototype = Tooltip.oldPrototype;
}
...
```

ChrisJPhoenix:
I've inserted your changes into the test and shared files. Thanks!

I hadn't realized you could pass the element, not just its ID, into the el field for new AceEditor. That certainly makes it easier.

I've also taken the liberty of deleting the faux_editor div from tests/tooltips/index.html.

And I've added color-picker support! It seems to work properly. I had to hack up external/colorpicker/colorpicker.js a bit. I only changed the __Hue and __Selector functions since it seems that's all we're using. I'm not 100% happy with passing strings rather than enum for start/mid/stop-scrub mode, but I don't know where to put the enum definition.

I didn't yet add the requested feature to put the correct tool tip back after an undo. Comparing between scrub-then-undo on fill(1,2,3); vs. rect(1,2,3,4); vs. foo(1); I see that the autosuggest only kicks in on the rect() function. On the other functions, the scrubber or color-picker come back as expected after undo. TBH, I find the autosuggest overly-aggressive on rect(); if I've got a syntactically correct rect() with numbers in place, and I click on a number when the cursor was elsewhere, I expect the scrubber, and I get the autosuggest. If the autosuggest were less aggressive on rect in general, then I think undo would put the correct tooltip in place on rect as it does on the other functions.

kevinb7:
> And I've added color-picker support! It seems to work properly. I had to hack up external/colorpicker/colorpicker.js a bit.

That's totally understandable considering the feature that you're implementing.

> I'm not 100% happy with passing strings rather than enum for start/mid/stop-scrub mode, but I don't know where to put the enum definition.

It probably makes sense it put it in tooltip-engine.js as a global.  I have a plan to clean up globals, but for now this should be fine.

> TBH, I find the autosuggest overly-aggressive on rect(); if I've got a syntactically correct rect() with numbers in place, and I click on a number when the cursor was elsewhere, I expect the scrubber, and I get the autosuggest.

This has been fixed in https://github.com/Khan/live-editor/commit/2ca622100ee81391b31671780b44e1e515110122.

ChrisJPhoenix:
I tried to merge 2ca6221 to see whether this fixes the tooltip-after-undo, and git keeps refusing.
```
F:\khan\live-editor-fork>git fetch

F:\khan\live-editor-fork>git merge 2ca622100ee
merge: 2ca622100ee - not something we can merge

F:\khan\live-editor-fork>git merge 2ca622100ee81391b31671780b44e1e515110122
merge: 2ca622100ee81391b31671780b44e1e515110122 - not something we can merge

F:\khan\live-editor-fork>git cherry-pick 2ca622100ee81391b31671780b44e1e51511012
2
fatal: bad object 2ca622100ee81391b31671780b44e1e515110122

F:\khan\live-editor-fork>git merge origin 2ca622100ee81391b31671780b44e1e5151101
22
merge: 2ca622100ee81391b31671780b44e1e515110122 - not something we can merge

F:\khan\live-editor-fork>git cherry-pick origin 2ca622100ee81391b31671780b44e1e5
15110122
fatal: bad object 2ca622100ee81391b31671780b44e1e515110122
```

20 minutes of googling didn't find how to get around this. What command should I be using? (I've been using git for less than two years - thus I'm still a novice. :-(  http://stevebennett.me/2012/02/24/10-things-i-hate-about-git/ and if more evidence were needed, http://sethrobertson.github.io/GitFixUm/fixup.html )

So, does the autosuggest fix also fix the wrong-tooltip-on-undo feature?

What are my next steps on this, or is it done?

kevinb7:
@ChrisJPhoenix this is why branches come in handy, but don't worry you can move your changes over to a branch.  

Start by create a new branch for your feature using tip of master: `git checkout -b avoid_undo_avalanche`.  Push this branch to origin: `git push origin avoid_undo_avalanche`.  Verify that this branch is on github with all of your changes because in the next step we're going to rewrite `master` and we don't want to destroy your changes.

Next do a hard reset on head, this will remove commits from master: `git checkout master` followed by `git reset --hard 8b516d41ea84e1f84b4e8c856314a585e182bfa3`.  Push the changes to origin via `git push origin master --force`.  The reason why `--force` is necessary is b/c you're rewriting master's history.  You'll have to eventually create a new pull request that diffs your `avoid_undo_avalanche` branch against our `master.

Okay, now you're ready to update master with changes from this repo.  You'll need to add need to add another remote, namely this repo, using `git remote add upstream https://github.com/Khan/live-editor`.  Note: `upstream` could be any name, just not `origin` because you already have `origin` set up.  You can can that this succeeded by doing `git remote -v` which lists all of your remote repos.

Now let's fetch the changes from `upstream` with `git fetch upstream` and then merge them in your local master via `git merge upstream/master`.  There should be no merge conflicts here.

Time to rebase.  This is where you'll run into conflicts.  Start by switching to `avoid_undo_avalanche` by running `git checkout avoid_undo_avalanche`.  Before merging, let's squash your commits into a single commit.  I find this makes merges easier b/c you'll only have one set of merge conflicts.  If you don't do this you may get merge conflicts for each of commits your in the branch you're trying to rebase which can become tedious to deal with.

To rebase run `git rebase -i HEAD~5`.  This will begin an interactive rebase of the last 5 commits in your the current branch.  It will open your favorite editor and you'll have to edit the message to specify which commits you want squashed.  Initial each commit will say "pick" beside it.  You want to keep the first one as "pick" and change the rest to "squash".  I should look something like this:
```
pick 3498cf0 Avoid undo avalanche on number-scrubbing
squash 9a08b90 Flaky undo tests
squash c8a4230 Creating a new editor messes up the "General -> Remove succeeds" test
squash 1cd3732 First-draft test of anti-undo hack - doesn't work yet.
squash 4246140 Anti-undo hack works for number scrubber and color picker. Tests are …

# ...
```
After rebasing you'll be left with a single commit in `avoid_undo_avalanche` with all of your changes.

Now to rebase your branch on top of `master` by running `git rebase master`.  It will "replay" all of your changes in `avoid_undo_avalanche` on top of all of the changes in `master`.  If you run into merge conflicts git will give you a chance to fix them.  After you fix them stage those files using `git add`.  Once all of the conflicts have been fixed and staged run `git rebase --continue`.  You should be left with a single commit in `avoid_undo_avalanche` with all of the changes from upstream master underneath it.  Create a new pull request between your `avoid_undo_avalanche` branch and our `master` branch.

ChrisJPhoenix:
I was able to do these steps with only a few problems which I could solve. Many thanks for detailed, clear, correct instructions. (I'd left gulp running, which caused merge conflicts in build/ halfway through, and I did too broad a git add and had to walk it back when it included webstorm stuff.)

Git managed to mess up whitespace or something in the colorpicker.js file. I tried to fix it, and wasn't able to in the time I have this morning. The actual changes are just a few lines, around lines 81, 106, 166, 171, 182, 187, 201, 206, 220, 225.

...And now gulp is refusing to run so I can't check my work.
```
F:\khan\live-editor-fork>gulp
module.js:338
    throw err;
          ^
Error: Cannot find module 'gulp-util'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:278:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (F:\khan\live-editor-fork\gulpfile.js:19:13)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)

F:\khan\live-editor-fork>
```

Sorry to leave it hanging here - I'll probably be able to do some work mid-week next week.

kevinb7:
The reason why gulp stopped working is that some of the new commits in master need new modules so you'll have to do an `npm install` again, sorry for not mentioning that.  

The whitespace in colorpicker.js should be spaces not tabs, I've had trouble with this file in the past always flipping between tabs and spaces so I changed it to spaces recently.  

In order to avoid including WebStorm stuff you can add the .idea folder to your global .gitignore.  Here's some info on how to set that up: http://stackoverflow.com/questions/7335420/global-git-ignore.

> Sorry to leave it hanging here - I'll probably be able to do some work mid-week next week.

No worries.  Whenever you have time.

ChrisJPhoenix:
OK, I replaced the tabs with spaces and re-pushed. Alas, I forgot to remove the build from the commit, and while I'm sure there's a good way to undo that in only 45 git commands, I can't afford to figure it out tonight. :-)

Also, 
1) When there's an "Oh Noes" in the code, for example
```
fill(128, 64, 128);
rect(206,1,45,9);
a(42);
```
 - a isn't defined - then the scrubber and color-picker cause the fill/rect code to be re-evaluated, but keystrokes (including Ctrl-Z) don't cause the code to be re-evaluated.

2) Even though the scrubber appears when I click on a number in rect(), it doesn't appear after undoing a scrub in rect(). And when I move the cursor with arrow keys, autosuggest comes up. Looking again at 2ca6221, it looks like you've got colorPicker and numberScrubberClick before autoSuggest in the ace_pjs list, but numberScrubber after autoSuggest. Shouldn't numberScrubber be before autoSuggest? 

If numberScrubber really should be after autoSuggest, then the only way I can see to make the scrubber appear after undo is to set a flag when the last editor event was an undo, and check that flag in the number-scrubber-click.js detector() function (which I think should be called after the undo). This seems too ugly/hacky, especially since it depends on _un_setting the flag on any editor event that wasn't an undo. In addition, to set and unset the flag, we either have to hack into the calls going into the editor - which I think really goes against the architecture of the code - or depend on the relative timing of the "undo happened" (and "something other than undo happened") vs. "tooltip requested" events.

kevinb7:
> I forgot to remove the build from the commit, and while I'm sure there's a good way to undo that in only 45 git commands

We want the build files to be committed.  If you want to change your last commit, make your changes and do `git commit -a --amend`.  This will amend all changes to your last commit.  You'll have to `git push origin <branch_name> --force` to push it to github though.

> a isn't defined - then the scrubber and color-picker cause the fill/rect code to be re-evaluated, but keystrokes (including Ctrl-Z) don't cause the code to be re-evaluated

Interesting.  Once you create a pull request and I'll check it out.

> Shouldn't numberScrubber be before autoSuggest?

That's what I thought originally.  There was a bug where the numberScrubber wasn't popping up at all.  So then I switch the order, but then it was hard to get the autoSuggest to pop up.  The numberScrubber beats autoSuggest if it's a click.

> This seems too ugly/hacky, especially since it depends on _un_setting the flag on any editor event that wasn't an undo. In addition, to set and unset the flag, we either have to hack into the calls going into the editor - which I think really goes against the architecture of the code - or depend on the relative timing of the "undo happened" (and "something other than undo happened") vs. "tooltip requested" events.

Let's leave forcing the numberScrubber to appear after an undo out of these changes.  It sounds like we need to some re-architecting around undo/redo and state.  We should probably have our own undo/redo manager that manages state of the tooltip widgets.

ChrisJPhoenix:
OK, I've created a pull request. I think it's what you want.

PR #398, opened 2015-08-24
Fixed failing test

Gigabyte-Giant:
...since the `expectedCode` was putting object properties on new lines, it was causing the test to fail.

I just put each object property on a new line, and the test seems to pass.


kevinb7:
Thanks!

PR #416, opened 2015-09-04
Update processing.js dep

jaredly:
Now including an LRU cache for fonts, to fix a memory leak in firefox.

I just did a `git checkout origin/master` in the processing dep, and then a `npm run build`.

Test Plan:
See pull request to processing-js for detailed info
https://github.com/Khan/processing-js/pull/8

### DIFF
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -1387,6 +1387,94 @@
     processingInstances.push(processing);
   };
 
+  ////////////////////////////////////////////////////////////////////////////
+  // LRUCache.JS START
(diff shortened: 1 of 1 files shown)


kevinb7:
@jaredly you should also run `gulp test_output_pjs`.  There should only be a single failure in the last 5 or so tests.  This is expected.  The test fails because an expected exception occurs asynchronously and mocha doesn't like that.

jaredly:
Ok, yeah -- saw that one failure and nothing else.

kevinb7:
@jaredly cool.  Thanks for fixing that issue.  ^_^

Issue #348, opened 2015-07-06
automate building and committing of compiled code that passes tests


kevinb7:
Even though ```gulp``` now does a rebuild the first time it's run, it's still possible to commit code in js/ that's out of sync with build/.  For example, if you make a really quick change to a source file and forget to run ```gulp``` then commit that change things will be out of sync.

This will probably take some doing to implement so in the meantime one of the things all code reviewers should check for is whether or not build files were commit or not.  If there are any changes to js/ then there should be changes to at least one file in build/.

kevinb7:
Another option that is maybe easier is to fail the build if there are changes in build files when travis-ci runs `gulp build`.

PR #453, opened 2015-10-17
Refactor test runners so that individual test suites can be run one at a time

kevinbarabash:
Summary:
In order to improve test reliability when running from the command line and
especially when running on travis-ci we'd like to be able to run fewer tests
at time before using phantomjs.  The theory is that fewer tests will require
less memory to run.  At the end of each suite we restart phantomjs in order
to clear the memory we've accumulated on the heap.

Test Plan:
- npm test
- create a pull request to check travis-ci integration

Reviwers: pamela john


js/output/pjs/pjs-code-injector.js
```diff
@@ -80,8 +80,9 @@ class PJSCodeInjector {
             this.initializeProps();
         }
 
+        // TODO(kevinb) add options for the initial timeout and frame timeout
         this.loopProtector = new LoopProtector(
-            infiniteLoopCallback, 2000, 500, true);
+            infiniteLoopCallback, 1000, 500, true);
```
kevinbarabash: The tests are looking stable after changing this to 1000.  The issues was that 2000ms is the same timeout used by mocha so sometimes the mocha timeout occurred first and sometimes it didn't.

tests/output/pjs/output_test.js
```diff
@@ -145,7 +145,7 @@ describe("Scratchpad Output Exec", function() {
     });
 
     failingTest("Too Many Draw Operations", function() {
-        for (var i = 0; i < 17000; i++) {
+        for (var i = 0; i < 1000000; i++) {
```
kevinbarabash: When testing locally 17000 wasn't enough to trigger the loop protector so this test was intermittently failing on travis-ci b/c although the loop protect code was being triggered, mocha was timing out b/c both timeouts were 2000ms.  This is enough to trigger it in both places.

kevinbarabash:
This partially solves https://github.com/Khan/live-editor/issues/348.  Eventually all tests should pass on travis-ci, but this is a step in the right direction.

kevinbarabash:
I've landed this using phabricator.

Issue #457, opened 2015-10-23
UX issues with tooltips


alexristich:
Four issues:

1) I noticed a flicker when clicking from within the body of one function to anywhere else in a program.

For example:
```
point(clientX, clientY);
image(getImage("avatars/leaf-blue"), 0, 0, 200, 200);
```

Say my cursor is between the "g" and "e" of `getImage()`. If I then click onto one of the parameters of `point()` then the tooltip for `image()` will flicker briefly before disappearing.

2) When the cursor is within the body of a function where tooltips are available (and the tooltip is showing), if I click onto a commented line of code, the tooltip doesn't disappear.

3) When writing a function which takes another function as a parameter (such as `image()` above), once I close the parens for the first function, the tooltip does not show for the rest of the function.

4) When typing on a commented line, the autosuggest feature remains active

Tracking down the first one took a while, as I'm new to the project so a lot of it is just getting used to the structure. I think I've figured out why it's happening, but I have a question.

```
        setTimeout(function(){
            if (shiftPressed) {
                editor.selection.selectToPosition(pos);
            }
            else if (!this.$clickSelection) {
                editor.selection.moveToPosition(pos);
            }
            this.select();
        }.bind(this), 0);
```

Is there any reason for the `setTimeout()` here?

Secondly, I noticed some inconsistencies when using the mouse and keyboard. With the keyboard, if I press left or right, the tooltip appears where the cursor moves to. If I click to the left or right with the mouse, the tooltip does not appear, and only appears on the second click on in that same location. Is the desire to have the tooltip show up right away when clicking with the mouse as well?

Anyway, I'm happy to continue to spend some time on these and will submit a PR with the four fixes once I've managed to get everything figured out.

kevinbarabash:
I'll have a look at it over the weekend.  You could try it without the `setTimeout` and see what happens.

alexristich:
Sure thing. I ask actually because I did try it, and removing `setTimeout` in this case fixes the flicker issue noted in item 1. It also results in a more consistent behavior between the mouse and keyboard, and I think a better user experience as well.

I ran the tests and there seems to be no difference in tests passing/failing when removing this, and couldn't find any cases where anything was breaking any other UI elements when testing through the `simple` demo. However, since I don't know why this code was added in the first place, I hesitate to remove it without asking any questions!

Also, sorry! I should have been more clear as to where this code is. Otherwise it's not too helpful :) This snippet is from editor_ace_deps.js:2400.

[...]

PR #462, opened 2015-10-26
Fixes to tooltip UI #457

alexristich:
Fixes for tooltip issues (issue #457). Includes:
1) Fix for tooltip flicker when moving the cursor with the mouse
2) Fix for tooltip not disappearing properly when moving from a function to a comment
3) Fix for autosuggest incorrectly triggering within comment


kevinbarabash:
You should be making changes to the files in js/ and then build the files using `gulp`.  I'll checkout the revision to try out the new behavior later this week when I have a bit more time.

alexristich:
Thanks for the heads up! I'll do that from now on. I still have a few more things to fix wrt the tooltips/popups and should hopefully have some time later this week or this weekend to spend on this.

alexristich:
I noticed one of my changes here introduced a new bug and ended up not being necessary. The bug was that the number-scrubber wasn't showing up properly for numbers within a function's parameters. I've just made another commit that reverts this fix, so it should be working properly again in this pending PR.

kevinbarabash:
I finally got around to checking out the code.  It definitely fixes some of the issues, but not all of them.  I have verified that the flickering is gone and the issue with the tooltip remaining up when the cursor is in a comment is gone.  The regex is still not quite right, but I feel like creating a regex that handles all situations is probably not possible.  Doing so will probably require inspecting the AST, but this should not be part of another diff.

Could you squash you commits and rebase?  It seems like there's some changes in the build files that are unrelated to the changes in the source files.

alexristich:
Duly noted. My first commit to the /build directory really muddled things up. Let me tidy things up hopefully this evening so it'll be ready to go.

On the regex, I definitely cheated somewhat - it indeed is very tricky to come up with something that covers all cases, and perhaps not a huge priority at this time. This patch will help from the user's perspective, but still leaves something to be desired.

In any case, will let you know when this PR should be ready to go.

alexristich:
Kevin, I've squashed the commit - took me a bit to figure it out, though hopefully I did it correctly. If not please let me know! Also, is there a better way to merge from the Khan/master branch, or is how I've done it here correct?

Regarding the flicker, the issue is in the `ace.js` file, which is from the `ace_builds` repo as you probably already know. The change required to fix the flicker needs to be made there, as the only way to fix it here is to make the change in the `/build` directory, which will just get overwritten on the next build.

I found making the following change fixed the flicker, and as far as I could tell it didn't cause any other undesirable side-effects:

```
setTimeout(function(){
            if (shiftPressed) {
                editor.selection.selectToPosition(pos);
            }
            else if (!this.$clickSelection) {
                editor.selection.moveToPosition(pos);
            }
            this.select();
        }.bind(this), 0);
```

to:
```if (shiftPressed) {
            editor.selection.selectToPosition(pos);
        }
        else if (!this.$clickSelection) {
            editor.selection.moveToPosition(pos);
        }
        this.select();
```

kevinbarabash:
> Regarding the flicker, the issue is in the ace.js file, which is from the ace_builds repo as you probably already know. The change required to fix the flicker needs to be made there, as the only way to fix it here is to make the change in the /build directory, which will just get overwritten on the next build.

I didn't realize that this required a change to ace.  We should try to make this change upstream and then update ace.

kevinbarabash:
Let's focus on getting the fixes for issues 2 and 3 merged in.  There are still a couple of things wrong with the pull request: it shouldn't include that first merge commit and the second commit should include the build files.  It's probably easiest to create a new pull request using a new branch similar to what you did with https://github.com/Khan/live-editor/pull/483.

alexristich:
Sounds good - I'll submit a new PR tonight for this one.

Issue #466, opened 2015-10-29
Image picker not updating properly after changing image


alexristich:
When you use `getImage` it triggers the image picker to appear. Hovering over reveals a set of images, and clicking on an image updates the text in the code to refer to that image. However, when you move the mouse away (`mouseleave`), the small popup still shows the previous image.

I made a quick fix that hides it altogether on the `mouseleave` event, but I imagine the preferred behavior is to have the updated image show in the small tooltip (which matches the behavior of the parameter tooltip). 

In any case, here's what it currently looks like: http://screencast.com/t/RJDom3Z3RaA

I'll make some tweaks soon hopefully to have it update the `currentTooltip` to reflect the newly selected image.

kevinbarabash:
I've noticed that too, but never reported it.  Thanks for opening an issue.

> I'll make some tweaks soon hopefully to have it update the currentTooltip to reflect the newly selected image.

Looking forward to it.

alexristich:
A little surprise... the fix is already ready :) Freshly pushed to my fork. If you don't mind, I'll wait to submit another PR until I've fixed that other pesky tooltip-related bug (hopefully later this week).

kevinbarabash:
Sounds good.  For future PRs, you may want to try using branches to avoid unnecessary dependencies.

alexristich:
Ah, will do that as well! A bit new to all this stuff, so I appreciate the tips as I get more familiar with it. I see my latest push got joined with the same PR, so I'll go ahead and create a branch this week where I'll push some more updates, then submit a PR from that branch.

kevinbarabash:
> but I imagine the preferred behavior is to have the updated image show in the small tooltip (which matches the behavior of the parameter tooltip).

I think this would be preferable.  Sorry for not saying so earlier.

PR #483, opened 2015-11-08
Fix for issue #466

alexristich:
Includes a fix for the issue with the image-picker as per issue #466.

### DIFF
--- a/build/workers/pjs/test-worker.js
+++ b/build/workers/pjs/test-worker.js
@@ -1,12 +1,15 @@
 // Mock out $._, since we don't use any of the sprintf functionality
+var i18n = {};
+i18n._ = function(str) { return str; };
+// TODO(kevinb) remove when all challenge test code is updated to use i18n._
 var $ = {};
-$._ = function(str) { return str;};
+$._ = i18n._;
 
 // We set window to self so that StructuredJS can find Esprima and
 // Underscore
-if (typeof window === "undefined") { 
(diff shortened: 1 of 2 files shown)


kevinbarabash:
I checked out the code and gave it a try.  Instead of hiding the small version of the tooltip on `mouseleave`, it be better to update it.  Updating the click handler like so should do the trick:
```
.on("click", ".image", function() {
                $(this).parents(".mediapicker").find(".active").removeClass("active");
                $(this).addClass("active");
                var path = $(this).attr("data-path");
                self.updateText(path);
                self.updateTooltip(`"${path}"`);
            })
```

kevinbarabash:
Also, you should include the build files with the changes.

alexristich:
Ah! I had tried that first but kept running into the problem where the `data-path` wasn't being passed properly, and couldn't figure out why that was happening. My solution here was less than ideal. Thanks for the tip on this. I'll submit another PR right now with this suggestion along with the relevant build file.

kevinbarabash:
Thanks.  Also, no need to create a new PR.  In this case you could've made the changes on your branch and committed them using `--amend` and then did a force push to github.

alexristich:
Good to know! I deleted the commit and was a bit surprised to see that it also closed the PR automatically, though in retrospect the behavior makes sense. In any case, your tidbits of advice are greatly appreciated.

Issue #485, opened 2015-11-09
Cursor type for number-scrubber isn't being updated properly


alexristich:
For the number-scrubber, when the user hovers over the left or the right arrow, the cursor should be changed to a pointer, but it stays as the `ew-resize` cursor.

The issue here is that the `tooltips.css` file only contains a relevant attribute selector in the case of structured-blocks. A new PR is coming momentarily with an added attribute selector for the standard case.

kevinbarabash:
@alexristich can you post screenshots for UI bugs?

alexristich:
Absolutely. Here's a quick screencast:
http://screencast.com/t/C2zBNVyNpyc

A screenshot is tricky for this one, since screenshots by default hide the cursor, though the screencast can be viewed in your browser and should give you a good sense of what this is referring to.

By the way, is the `.block` CSS class in `tooltips.css` actually being used at the moment?

kevinbarabash:
> By the way, is the `.block` CSS class in `tooltips.css` actually being used at the moment?

It doesn't appear to be.

kevinbarabash:
Closed by https://github.com/Khan/live-editor/pull/486.

alexristich:
Good to know. I'll work around it in the meantime - the SVG wrt the number scrubber icons is next on my list.

PR #486, opened 2015-11-09
Fix for number-scrubber cursor issue as per #485

alexristich:
Added a couple lines to the CSS file to properly update the cursor when hovering over the number-scrubber.


kevinbarabash:
Looks good.  Thanks for the PR.

Issue #445, opened 2015-10-15
KA Javascript is now a distinct (and wrong) dialect


lgto4:
Refer to the simple demo program at https://www.khanacademy.org/computer-programming/wat/5726354510315520 .

The program declares var mouseClciked = inside of the draw function.  Being a local variable, the assignment should have no affect on Processing.js.  Alas, through a rewrite bug the local variable assignment is magically hoisted to a global variable assignment.

This is very bad.  Beginners need to learn when and how to use "var".  And, the language at KA is no longer Javascript.

kevinbarabash:
@lgto4 agreed.  This is a fixable problem, but there are many existing programs have have code that does this.  I think that we should first provide warnings against this.  After the warnings have been live on the site for a while then we can change the behavior.  It might also help to have some sort of common mistakes video on the site too.  @pamelafox thoughts?

lgto4:
@kevinbarabash: I should hope that the bug is fixable.  There can't be that many programs that have been led astray since the bug was injected with the most recent rendition of the KA infinite loop detector. Furthermore, breakage has never stopped the "forward" progress of KA features. It certainly should not stop the correction of an insidious bug.

kevinbarabash:
@lgto4 I thought this was always the behavior.  How do you know that this bug started with the loop detector?

kevinbarabash:
@lgto4 after some investigation, I think you're right about when it was introduced.  This bug should be pretty easy to fix.  I'll try to get to it sometime this week.

lgto4:
@kevinbarabash: Thanks for your attention to this issie.  I apologize for sitting on it. I am pretty clueless when it comes to dealing with github...

kevinbarabash:
@lgto4 no worries.  If you have an github related questions, ask away.  There's also a gitter chat room for this project: https://gitter.im/Khan/live-editor.

Potato42:
This is an extremely easy to fix bug.  Basically only one line needs to be removed: https://github.com/Khan/live-editor/blob/master/js/output/pjs/pjs-ast-transforms.js#L127.  I believe this should do it https://github.com/Khan/live-editor/pull/447.

However, this will break at least a few programs.

Potato42:
Wait, never mind, that doesn't seem to work.  -_-

JavascriptFTW:
[...]

PR #489, opened 2015-11-09
Fix improper handling of draw loop methods

JavascriptFTW:
This pull request fixes improper handling of the draw loop methods reported in #445 and modifies the flawed tests that were checking this to be correct now.

# Major Changes

## js/output/pjs/pjs-ast-transform.js

* Delete some AST transform code that was causing improper behavior of draw loop methods in local scopes.
* Edit comments to help make sure that no one else makes this same mistake in the future.

# Minor Changes

## tests/output/pjs/ast_transform_test.js

* Edit flawed tests to test for the proper output

Thank you for your time.

### DIFF
--- a/js/output/pjs/pjs-ast-transforms.js
+++ b/js/output/pjs/pjs-ast-transforms.js
@@ -120,11 +120,10 @@ ASTTransforms.rewriteContextVariables = function(envName, context) {
                     }
 
                     // Prefix identifiers that exist in the context object and
-                    // have not been defined in any scope or are one of the draw
-                    // loop functions.  Also, prefix any other identifers that
+                    // have not been defined in any scope.
+                    // Also, prefix any other identifers that
                     // exist at the global scope.
                     if ((node.name in context && scopeIndex === -1) ||
-                            drawLoopMethods.includes(node.name) ||
                             scopeIndex === 0) {
                         return b.MemberExpression(
                             b.Identifier(envName), b.Identifier(node.name));
@@ -146,9 +145,9 @@ ASTTransforms.rewriteContextVariables = function(envName, context) {
 
                     // Rewrite all function declarations, e.g.
                     // var foo = function () {} => __env__.foo = function () {}
-                    // that appear in the global scope unless it's one of the draw loop
-                    // methods.  In that case, always rewrite it.
-                    if (scopes.length === 1 || drawLoopMethods.includes(decl.id.name)) {
+                    // that appear in the global scope. Draw loop methods aren't
+                    // special, they should be treated in the exact same way.
+                    if (scopes.length === 1) {
                         if (["Program", "BlockStatement", "SwitchCase"].includes(parent.type)) {
                             return b.ExpressionStatement(
                                 b.AssignmentExpression(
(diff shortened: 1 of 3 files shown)


build/js/live-editor.editor_structured_blocks_deps.js
```diff
@@ -8786,7 +8786,7 @@ var JSComment = JSRules.addRule(JSRule.extend({
         "input input": "onInput"
     },
 
-    defaultValue: i18n._("Your comment..."),
+    defaultValue: $._("Your comment..."),
```
kevinbarabash: This should not be reverted.

external/structured-blocks
```diff
@@ -1 +1 @@
-Subproject commit 53ff5026034c4db53872d625dd2191f51e915426
+Subproject commit d3c8e897e60d664121c5956077a1204b3220ca16
```
kevinbarabash: This should not be reverted either.

js/output/pjs/pjs-ast-transforms.js
```diff
@@ -223,24 +223,20 @@ ASTTransforms.rewriteContextVariables = function(envName, context) {
...
+                        //     var y = 10;
+                        // };
+                        // It should NOT convert them to something like this:
+                        // __env__.draw = function() {
+                        //     var x = 5;
                         //     __env__.mouseClicked = function () { ... };
                         //     var y = 10;
                         // };
-
+
```
kevinbarabash: Including what the code shouldn't do isn't necessary and just adds noise.

kevinbarabash:
Looks pretty good except for a few things.  You should be able to fix things by doing:
- `cd external/structured-blocks`
- `git checkout 53ff5026034c4db53872d625dd2191f51e915426` (this will checkout the correct version)
- run `gulp` in the `live-editor` folder

This should get the things that accidentally got reverted (see my comments on the diff).

I think what happened is that you update masted without updating the submodule.  Whenever you update a branch (or checkout another branch), run `git submodule update` to make sure the submodules are in the right place.

JavascriptFTW:
Thanks for the tip @kevinbarabash will do that right now. Will also rebase after I have.

kevinbarabash:
There's one failure in the `test_output_pjs_ast_transform` suite that's my fault.  If everything else passes, squash the commits and this will be good to go.

JavascriptFTW:
@kevinbarabash I've gotten all tests but two to work. One is behaving differently in the test environment than in my local environment from what I can tell, and the other I'm pretty sure I have nothing to do with. I'm going to turn in for the night and try fixing these tomorrow.

JavascriptFTW:
@kevinbarabash I've been trying to fix the failing tests, but I haven't been able to see anything that could be wrong with my test (or code). I've `console.log`'d the transformed code on the demo page (my old fashioned way of testing since I can't run `gulp test`) and it matches what I'd expect the output to be.

A friend who can run `gulp test` volunteered to try and see what was going wrong with my fork and they ran into this:

http://take.ms/HAeE9

Do you have any ideas what might be going wrong?

Sorry that this is turning out to be more involved than it should be.

kevinbarabash:
> Do you have any ideas what might be going wrong?

I fixed that issue in https://github.com/Khan/live-editor/commit/65fff9f408f260244cf84f7b7cc57c77757dfbc0.  I was using ES6 syntax by accident (lines 238 and 247) in the test code.  I hadn't noticed because the build wasn't failing.

You can also run the tests in the browser by opening `tests/output/pjs/index.html`.  To run a specific test suite open `test/output/pjs/index.html?tests=ast_transform_test.js`.  Make sure you load these files from a local web server.

I find debugging issues much easier in the browser because you can set breakpoints.  Sometimes though the issues isn't reproducible in the browser.  Thankfully, in this case, the failures we're seeing on travis-ci do show up when running in the browser.

JavascriptFTW:
I believe I've found my problem. The test code was using a custom object for the global scope, of which `mouseClicked` was not a part. The test was expecting `mouseClicked` to be a part of the global scope. Well at least I won't make this mistake again :)

I'm going to add `mouseClicked` and that will most likely fix the issue on travis-ci.

The PR should be ready for merging in just a few after I run the tests (and squash of course).

Thanks for the help @kevinbarabash!

JavascriptFTW:
Except for one failure in pjs_ast_transform (which I believe is the one you said was okay) my code passes all tests. I'll rebase when you're ready to merge.

Thanks for your time @kevinbarabash!

kevinbarabash:
@JavascriptFTW everything looks good now.  I'm ready to merge.  Good work debugging.  I'll try to fix that failing test tonight so we can rely on travis-ci more.

JavascriptFTW:
@kevinbarabash Squashed and ready to merge sir!

kevinbarabash:
@JavascriptFTW awesome possum.  Thanks again for the pull request.

PR #442, opened 2015-10-12
Separate AST Transform to rewrite NewExpressions

Gigabyte-Giant:
This PR contains the changes that @kevinbarabash suggested to me to fix #324.
Currently, the live-editor performs a string replacement to replace any and all instances of *anything* that matches the following Regex:
```
/\bnew[\s\n]+([A-Z]{1,2}[a-zA-Z0-9_]+)([\s\n]*\()/g
```

Anything that matches that Regex is replaced with the following:

```
PJSOutput.applyInstance($1,'$1')$2");
```

In the end, the following code...
```javascript
var Obj = function(prop) {
    this.prop = prop;
};

var myObj = new Obj(10);
```
...gets changed to...
```javascript
__env__.Obj = function(prop) {
    this.prop = prop;
};

__env__.myObj = __env__.PJSOutput.applyInstance(__env__.Obj, 'Obj')(10);
```

The current approach would work fine, but, what about innocent strings? For example, see [this program](https://www.khanacademy.org/computer-programming/live-editor-bug-example/6383254823895040).

The problem is, the regular expression will also match things that are inside of strings.

This PR gets rid of the replacements, and instead, uses AST Transforms to transform all `NewExpression`s to `CallExpression`s.

The approach is quite different, **but the output is exactly the same**.

Notable changes:
 * I added an AST Transform that changes all `NewExpression`s with `CallExpression`s.
 * I ~~commented out~~ removed the string replacements that were originally used to achieve the same result.
 * I added an AST Transform test to make sure that the AST Transform worked correctly.

### A note to whoever reviews this:
Due to the fact that this is an entirely different approach than was taken before, some of the older tests fail (from what I can tell, it's only tests in `tests/output/pjs/output_test.js`)...
Can you tell me which tests are no longer valid? I've commented out one test that I found that might not be valid anymore (test title: "should transform constructors with multiple capital letters").

I'll squash my commits once I resolve the failing tests.

Gigabyte Giant (brynden.bielefeld@hotmail.com)

PR #490, opened 2015-11-10
Fix for issue #457

alexristich:
This includes fixes for items 2 and 3 as reported in issue #457.

Note that both 1 and 4 from the original report will need to be fixed upstream in the ace editor. I can submit a PR there if desired.

I also slightly tweaked the regex from the previous PR, though it still needs the `isInParenthesis` check to work properly. Not the prettiest thing ever, but it does the trick for now - at least until a better approach is implemented.

### DIFF
--- a/build/js/live-editor.tooltips.js
+++ b/build/js/live-editor.tooltips.js
@@ -1483,6 +1483,11 @@ window.TooltipEngine = Backbone.View.extend({
             return false;
         }
         if (this.isWithinComment(params.pre)) {
+            // if selected text is within a comment, hide current tooltip (if any) and return
+            if (this.currentTooltip) {
+                this.currentTooltip.$el.hide();
+                this.currentTooltip = undefined;
+            }
             return false;
         }
         this.last = params;
@@ -1676,7 +1681,13 @@ TooltipEngine.classes.autoSuggest = TooltipBase.extend({
     },
 
     detector: function detector(event) {
-        if (!/(\b[^\d\W][\w]*)\s*\(\s*([^\)]*)$/.test(event.pre) || this.parent.options.record.playing) {
+        // TODO: update this to support auto-suggest tooltip for inner functions passed as params
+        // this currently only allows displaying of the tooltip for the outside function, except in cases
+        // where the inner function uses one of the other tooltips (e.g. image-picker)
+        if (!/(\b[^\d\W][\w]*)\s*(\({1}\s*\){1})*\s*([^\]]*)$/.test(event.pre) || this.parent.options.record.playing) {
+            return;
+        }
+        if (!this.isInParenthesis(RegExp.$3)) {
             return;
         }
         if (event.source && event.source.type === "changeCursor" && this.mouse) {
@@ -1684,7 +1695,7 @@ TooltipEngine.classes.autoSuggest = TooltipBase.extend({
             return;
         }
         var functionCall = RegExp.$1;
-        var paramsToCursor = RegExp.$2;
+        var paramsToCursor = RegExp.$3;
         var lookupParams = ScratchpadAutosuggest.lookupParamsSafeHTML(functionCall, paramsToCursor);
         if (lookupParams) {
             this.aceLocation = {
--- a/js/ui/tooltip-engine.js
+++ b/js/ui/tooltip-engine.js
@@ -163,6 +163,11 @@ window.TooltipEngine = Backbone.View.extend({
             return false;
         }
         if (this.isWithinComment(params.pre)){
+            // if selected text is within a comment, hide current tooltip (if any) and return
+            if (this.currentTooltip) {
+                this.currentTooltip.$el.hide();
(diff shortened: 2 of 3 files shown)


alexristich:
Hmm, I see the Travis CI build failed here. I ran the tests on my local machine before pushing - two were failing (`"Scratchpad Output Exec Too Many Draw Operations"`), though these were failing before I made any changes and don't seem to be related to what this commit changed. Not sure why Travis CI didn't like this one.

Let me know if I need to do anything here to clean this one up.

Gigabyte-Giant:
Hi @alexristich!

Below is a bit of information regarding the failing tests:
 * "Scratchpad Output Exec Too Many Draw Operations"
  * According to @kevinbarabash (see the first line of [this comment](https://github.com/Khan/live-editor/pull/442#issuecomment-155113258)), those tests are expected to fail.

 * "AST Transforms for exporting should automatically loop when a 'draw' method exists"
  * Again, according to @kevinbarabash (see [this comment](https://github.com/Khan/live-editor/pull/489#issuecomment-155588331)), that test fails due to ECMAScript 6 syntax being used in the test code. He submitted a diff (https://github.com/Khan/live-editor/commit/65fff9f408f260244cf84f7b7cc57c77757dfbc0) that should solve that failing test.

kevinbarabash:
> "AST Transforms for exporting should automatically loop when a 'draw' method exists"

There's another reason for this failure.  I've submit a patch internally and it's been reviewed, I just haven't pushed it yet.

alexristich:
@Gigabyte-Giant @kevinbarabash Thanks for the clarification! In that case I'll leave this one as is.

build/js/live-editor.tooltips.js
```diff
@@ -1483,7 +1483,8 @@ window.TooltipEngine = Backbone.View.extend({
             return false;
         }
         if (this.isWithinComment(params.pre)) {
-            return false;
+            // if within comment, emit default event to hide existing tooltip
+            params = this.requestTooltipDefaultCallback;
```
kevinbarabash: What is this supposed to do?

alexristich: This fixes the issue where, when a tooltip is showing, if the user clicks on a commented line of code, the displayed tooltip doesn't disappear.
Returning false prevents another tooltip from being displayed, but doesn't do anything to hide the existing tooltip.

Originally I had two lines here, directly from the code which ends up being called when emitting this event (see lones 1433-1440).

Do you think it would be better instead to stick with the original idea and simply paste these lines in place of the line of code in question?

```
this.currentTooltip.$el.hide();
this.currentTooltip = undefined;
```

alexristich:
I just submitted another commit with an updated version. Let me know if this is better.

I don't quite know how to delete the older commit without closing the PR and having to create a new one. I tried using `git cherry-pick` but it resulted in a conflict. In any case, if this latest code looks good I can clear things out and resubmit this code from a new branch in a new PR (or, if you have a chance, it would be awesome if you might be able to point me to where I can learn more about how to do this).

kevinbarabash:
@alexristich thanks for making the change.  I was confused by setting params to be a callback function when it was a initially being set to a object.  This makes more sense.

As for getting rid of the older commit, you can squash commits using `git rebase -i HEAD~2`.  This will interactively rebase the last two commits.  See http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html for details.

After you've squash the commits you'll have to do `git push --force` because you're rewriting history.

alexristich:
Yeah, in retrospect I think I was just trying to do something clever there, but definitely not the right approach. Thanks for the code review!

Excellent - I've squashed the commits and overwritten the branch, so this PR should hopefully be good to go. Let me know how I ought to proceed with issues 1 and 4 relating to the ace editor.

kevinbarabash:
@alexristich  thanks for the PR.  In the future it's probably best have separate issues and separate pull requests for separate issues.  I thought that this PR fixed issue 4 as well (not showing tooltips in comments) because if the change you made inside the `if (this.isWithinComment(params.pre))` block.

alexristich:
Yes, you're right! Sorry for the confusion on this one. This first report and subsequent PR was a mess, so thanks for bearing with me as I get more up to speed on how things are done around here!

Just for clarification, issue 4 in the original list is related to the ace editor, so while it can be fixed by directly editing the build file, it will be overwritten on the next build since it draws from the ace editor dependency. That issue, specifically, is that the autosuggest list for function names appears when typing on a commented line. 

The issue that this fixed was that, when clicking onto a commented line of code, if a tooltip was showing up then it wouldn't disappear.

kevinbarabash:
@alexristich no worries, thanks for the clarification.  I forgot that there were two issues having to do with comments.

pamelafox:
I've deployed this PR, can you verify the fixes are working as expected?

alexristich:
@pamelafox Excellent! I checked and they're all working as expected on the production server.

PR #493, opened 2015-11-14
Make errors for using advanced JS features friendlier

### LINKED ISSUE #492: Unhelpful Error Message When Using ES6/7 Features
The error message thrown by jshint when you attempt to use ES6/7 features in the live editor is "'some cool feature' is only available in JavaScript 1.7.". While this is an acceptable message it may not be the most helpful for some students.

JavascriptFTW:
This pull request fixes #492 by editing the jshint messages for using ES7 and ES6 features to be worded in a more friendly and helpful way.

# Minor Changes

## external/jshint/jshint.js

* Switch W104 from "'some feature' is only available in JavaScript 1.7." to "It looks like you're trying to use 'some feature'. 'some feature' is a JavaScript 1.7 feature and is not currently supported in the live editor."

* Switch W109 from "'some feature' is only available in ES6 (use esnext option)." to "'some feature' is only available in ES6 which is not supported by the live editor at this point."

If any editing is required please let me know.

Thank you for your time.

### DIFF
--- a/build/external/jshint/jshint.js
+++ b/build/external/jshint/jshint.js
@@ -58488,7 +58488,8 @@ var warnings = {
 	W101: i18n._("Line is too long."),
 	W102: i18n._("Trailing whitespace."),
 	W103: i18n._("The '{a}' property is deprecated."),
-	W104: i18n._("'{a}' is only available in JavaScript 1.7."),
+	W104: i18n._("'{a}' is only available in ES6 which is not supported by " + 
+        "this environment."),
 	W105: i18n._("Unexpected {a} in '{b}'."),
 	W106: i18n._("Identifier '{a}' is not in camel case."),
 	W107: i18n._("Script URL."),
@@ -58503,7 +58504,8 @@ var warnings = {
 	W117: i18n._("\"{a}\" is not defined. Make sure you're spelling it correctly " +
 		"and that you declared it."),
 	W118: i18n._("'{a}' is only available in Mozilla JavaScript extensions (use moz option)."),
-	W119: i18n._("'{a}' is only available in ES6 (use esnext option)."),
+	W119: i18n._("'{a}' is only available in ES6 which is not supported by " + 
+        "this environment."),
 	W120: i18n._("You might be leaking a variable ({a}) here."),
 	W121: i18n._("Extending prototype of native object: '{a}'."),
 	W122: i18n._("Invalid typeof value '{a}'"),
--- a/external/jshint/jshint.js
+++ b/external/jshint/jshint.js
@@ -58488,7 +58488,8 @@ var warnings = {
 	W101: i18n._("Line is too long."),
 	W102: i18n._("Trailing whitespace."),
 	W103: i18n._("The '{a}' property is deprecated."),
-	W104: i18n._("'{a}' is only available in JavaScript 1.7."),
+	W104: i18n._("'{a}' is only available in ES6 which is not supported by " + 
+        "this environment."),
 	W105: i18n._("Unexpected {a} in '{b}'."),
 	W106: i18n._("Identifier '{a}' is not in camel case."),
 	W107: i18n._("Script URL."),
@@ -58503,7 +58504,8 @@ var warnings = {
 	W117: i18n._("\"{a}\" is not defined. Make sure you're spelling it correctly " +
 		"and that you declared it."),
 	W118: i18n._("'{a}' is only available in Mozilla JavaScript extensions (use moz option)."),
-	W119: i18n._("'{a}' is only available in ES6 (use esnext option)."),
+	W119: i18n._("'{a}' is only available in ES6 which is not supported by " + 
+        "this environment."),
 	W120: i18n._("You might be leaking a variable ({a}) here."),
 	W121: i18n._("Extending prototype of native object: '{a}'."),
 	W122: i18n._("Invalid typeof value '{a}'"),
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -78622,7 +78622,8 @@ var warnings = {
 	W101: i18n._("Line is too long."),
 	W102: i18n._("Trailing whitespace."),
 	W103: i18n._("The '{a}' property is deprecated."),
-	W104: i18n._("'{a}' is only available in JavaScript 1.7."),
+	W104: i18n._("'{a}' is only available in ES6 which is not supported by " + 
+        "this environment."),
 	W105: i18n._("Unexpected {a} in '{b}'."),
 	W106: i18n._("Identifier '{a}' is not in camel case."),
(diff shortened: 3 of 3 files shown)


build/external/jshint/jshint.js
```diff
@@ -58488,7 +58488,9 @@ var warnings = {
 	W101: i18n._("Line is too long."),
 	W102: i18n._("Trailing whitespace."),
 	W103: i18n._("The '{a}' property is deprecated."),
-	W104: i18n._("'{a}' is only available in JavaScript 1.7."),
+	W104: i18n._("It looks like you're trying to use '{a}'. '{a}' is a " +
+        "JavaScript 1.7 feature and is not currently supported in the live " +
```
pamelafox: Can you change "live editor" to "this editor" or "this environment"?

JavascriptFTW: Will do!

JavascriptFTW:
@pamelafox I've made the edit you suggested. Let me know if there's anything else I should do.

build/external/jshint/jshint.js
```diff
@@ -58503,7 +58505,8 @@ var warnings = {
 	W117: i18n._("\"{a}\" is not defined. Make sure you're spelling it correctly " +
 		"and that you declared it."),
 	W118: i18n._("'{a}' is only available in Mozilla JavaScript extensions (use moz option)."),
-	W119: i18n._("'{a}' is only available in ES6 (use esnext option)."),
+	W119: i18n._("'{a}' is only available in ES6 which is not supported by " +
```
pamelafox: Change to environment

external/jshint/jshint.js
```diff
@@ -58488,7 +58488,9 @@ var warnings = {
 	W101: i18n._("Line is too long."),
 	W102: i18n._("Trailing whitespace."),
 	W103: i18n._("The '{a}' property is deprecated."),
-	W104: i18n._("'{a}' is only available in JavaScript 1.7."),
+	W104: i18n._("It looks like you're trying to use '{a}'. '{a}' is a " +
+        "JavaScript 1.7 feature and is not currently supported in this " +
```
pamelafox: Forgive me, this might be me not being up to date on things - what's difference between JavaScript 1.7 and ES6, which parts of the spec actually made it to browsers? Ideally we'd say either JS or ES, not both, but maybe that's not possible.

JavascriptFTW: @pamelafox I think that ES6 is the older new version while JavaScript 1.7 is still in progress. I did some looking around and let, const, and destructuring assignments are all 1.7 features, while pretty much everything else new is ES6 (see [1.7](https://developer.mozilla.org/en-US/docs/Web/JavaScript/New_in_JavaScript/1.7) and [ES6](https://developer.mozilla.org/en-US/docs/Web/JavaScript/New_in_JavaScript/ECMAScript_6_support_in_Mozilla)). I'm pretty sure this could be changed if you think it's necessary, the only reason I differentiated is because the original messages differentiated between ES6 and JavaScript 1.7.

build/external/jshint/jshint.js
JavascriptFTW: Oops. Sorry bout that. I do not have access to my currently but once I do I'll fix this and let you know. Sorry for the goof-up.

GoToLoop:
> I did some looking around and `let`, `const`, and destructuring assignments are all 1.7 features, ...

* Actually all of them are ES6 a.K.a ES2015.
* And `const` was picked up so early by ES5 browsers that it should be considered ES5 in practice:
https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const
* `let` took a lot more though. And in order to use it, it demands `'use strict';`:
https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let
* ES7 instead is gonna bring us very advanced stuff like SIMD:
https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SIMD

IMO, Khan shouldn't bar ES6 features at all. Specially something as old as `const`!

kevinbarabash:
> IMO, Khan shouldn't bar ES6 features at all. Specially something as old as const!

It's not a question of barring ES6 features, it's a question of supporting them.  We want all CS programs to run cross browser which means that we'd have to babel to compile them.  We'd also have to update the linter (or switch to eslint if we want to support all ES6).

GoToLoop:
> We want all CS programs to run cross browser which means that we'd have to babel to compile them.

* That all depends on establishing the minimum browser version full support is guaranteed.
* Keyword `const` for example is recognized on very old browsers like Opera 12 and no need for Babel.
* However, while IE11 works, IE10 doesn't recognize `const`.
* But IE browser is gonna be defunct soon and be replaced by Edge.
* It's a pity to bar folks to use some ES6 features when it's already available in modern browsers.
* I bet classes gonna bring lotsa Java, C#, Python, etc. programmers to JS camp!  :D
https://developer.Mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

JavascriptFTW:
@pamelafox Sorry for that goof-up I've fixed it now and everything should be ready. Very sorry for that.

kevinbarabash:
> However, while IE11 works, IE10 doesn't recognize `const`.

While we're dropping IE9 support as of Jan 2016, we'll still be support IE10 for a while.

> I bet classes gonna bring lotsa Java, C#, Python, etc. programmers to JS camp! :D

They also make using classes less intimidating for beginner programmers.

kevinbarabash:
@JavascriptFTW instead of merging master can you rebase your changes on to master?

JavascriptFTW:
@kevinbarabash It's rebased now. It should be ready for merging once the tests are complete.

external/jshint/jshint.js
kevinbarabash: Can you change the message here to say ES6 as well?  You can `git commit --amend` to update the diff and then do a force push.  I'll merge after that.

JavascriptFTW: I'll edit this in a bit as I cannot currently. I said "JavaScript 1.7" instead of "ES6" as it matched the phrasing of the original message.

kevinbarabash: You could just use the same message as W119.

JavascriptFTW:
@kevinbarabash I believe that I've changed the message according to your suggestion.

kevinbarabash:
Excellent.  LGTM.

kevinbarabash:
@JavascriptFTW thanks for the pull request.

GoToLoop:
> While we're dropping IE9 support as of Jan 2016, we'll still be support IE10 for a while.

Seems like not even MS wanna support those old browsers next year:  :-D
https://www.Microsoft.com/en-us/WindowsForBusiness/End-of-IE-support

For now IE11 survives. But I bet it's not gonna take too long for MS to support Edge only...  @_@

Issue #496, opened 2015-11-16
Can't make a object instance when the object constructor is passed into a function


petercollingridge:
The following code gives an error: Cannot read property 'prototype' of undefined.

    var makeObj = function(obj) {
      return new obj();
    };

    var x = makeObj(Array);

It worked about 10 days ago. It's not a big deal, but I've seen a few programs broken as a result of it.

kevinbarabash:
This probably has to do with a recent change to rewrite `NewExpressions` using an AST transform as opposed to a regex.  @Gigabyte-Giant if you have some time could you look at this?

Gigabyte-Giant:
I was just looking over this @kevinbarabash. Give me a bit and I'll get back to you.

MikaalSky:
I'm not sure it's neccasary to make arrays like that but I can see for instance where you could put a Object in a parameter to create a new instance and append it but hardcoding it would make sense to do 99% of the time. Might as well get it fixed though.

Gigabyte-Giant:
Okay, I've determined what I *think* the problem is; [my AST transform](https://github.com/Khan/live-editor/blob/master/js/output/pjs/pjs-ast-transforms.js#L293-L319) **always** prefixes identifiers with `__env__`, which is fine, unless you're referencing a function parameter (or at least that's what I can tell from my quick test).

I broke it, so I'll fix it. I'll also write a a new test to make sure that won't happen again.

Sorry about that, guys!

kevinbarabash:
@Gigabyte-Giant this is a regression.  Thanks for fixing.

petercollingridge:
Wow, that was very quick, thank you!

@MikaalSky I just used `Array` as a simple example; none of the programs I saw were actually doing this, they were all using custom objects.

petercollingridge:
I think there's still a related issue, since it gives the same error and broke at around the same time.

The problem is from this program (which has been fixed by commenting out some parts): https://www.khanacademy.org/computer-programming/_/6380002294366208

The issue seems to be due to defining and creating an object inside a self-invoking function, like this:

    (function() {
        var f = function() {};
        var g = new f();
    })();

Again, it doesn't seem a massive problem, since hardly anyone is going to do this, but just to let you know.

Gigabyte-Giant:
Hi @petercollingridge! Thanks for providing me with another test case. I ran the code you provided, against my branch, and all appears to work just fine (in other words, my PR should resolve all of those issues).

[...]

PR #497, opened 2015-11-16
Resolve regression bug #496

Gigabyte-Giant:
https://github.com/Khan/live-editor/issues/496 was caused due to the fact that the AST transform created in https://github.com/Khan/live-editor/commit/cd8f6dc529f387ad107a610a947254ff5ddbdbb2, was always prefixing identifiers with `__env__`.
Function parameters are not prefixed with `__env__`, therefore trying to reference  `__env__.someFunctionParam` would inevitably fail.

This commit adds a ternary-if-statement that checks to see if we're inside of a `FunctionExpression`. If we are, we don't prefix identifiers when we rewrite `NewExpression`s to `CallExpression`s.

I also added a test to make sure that this change did what it was supposed to do.

Gigabyte Giant (brynden.bielefeld@hotmail.com)

### DIFF
--- a/js/output/pjs/pjs-ast-transforms.js
+++ b/js/output/pjs/pjs-ast-transforms.js
@@ -304,10 +304,7 @@ ASTTransforms.rewriteNewExpressions = function(envName) {
                             b.Identifier("applyInstance")
                         ),
                         [
-                            b.MemberExpression(
-                                b.Identifier(envName),
-                                b.Identifier(node.callee.name)
-                            ),
+                            node.callee,
                             b.Literal(node.callee.name)
                         ]
                     ),
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -87849,7 +87849,7 @@ ASTTransforms.rewriteNewExpressions = function (envName) {
     return {
         leave: function leave(node, path) {
             if (node.type === "NewExpression") {
-                return b.CallExpression(b.CallExpression(b.MemberExpression(b.MemberExpression(b.Identifier(envName), b.Identifier("PJSOutput")), b.Identifier("applyInstance")), [b.MemberExpression(b.Identifier(envName), b.Identifier(node.callee.name)), b.Literal(node.callee.name)]), node.arguments);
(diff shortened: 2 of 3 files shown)


js/output/pjs/pjs-ast-transforms.js
```diff
@@ -304,10 +304,15 @@ ASTTransforms.rewriteNewExpressions = function(envName) {
                             b.Identifier("applyInstance")
                         ),
                         [
+                            // If we're not currently inside of a function,
+                            //  then prefixing identifiers is alright.
+                            // However, if we are inside of a function,
+                            //  then we shouldn't be prefixing function params.
+                            (/^Function/.test(path[path.length - 4].type) === false ?
```
kevinbarabash: This seems unreliable.  What if you're inside a loop inside a function.  You should probably search for /^Function/ in all ancestor nodes.

kevinbarabash:
@Gigabyte-Giant could you write a test for the scenario I mentioned in my other comment.  Also, in terms of testing, not all tests need to look at the AST.  You could just write tests in same the style in output_test.js verifying that the code compiles and runs.  We should probably rewrite some of the existing AST tests too.  I think there's only a few where we can't verify correct behavior any other way.

Gigabyte-Giant:
Sounds good @kevinbarabash. I'll make the aforementioned change(s)...

I'd also be more than happy to help re-write/improve existing tests/code. If you need an extra set of hands, feel free to ping me on Gitter or by email.

Gigabyte-Giant:
@kevinbarabash actually, I never need to test for `/^Function/`. `ASTTransforms.rewriteNewExpressions` is run *after* `ASTTransforms.rewriteContextVariables`, therefore things that need to be prefixed, will already be prefixed in `node.callee`.

I'll be removing the ternary-if-statement. I can still add another test, if you'd like.

kevinbarabash:
That makes sense.  Can you squash this?

Gigabyte-Giant:
Done @kevinbarabash!

kevinbarabash:
@Gigabyte-Giant awesome, thanks for the quick response.

Gigabyte-Giant:
Happy to help! Sorry to break it in the first place.

PR #502, opened 2015-11-24
Replace CRLF line endings with LF endings

### LINKED ISSUE #333: AceEditor Edit to Fix some CR+LF Issues
Brought to my attention by @SpongeJr ...
CR + LF causes double-spacing on windows computers when copy-pasting ka code in another program and saving. If we could axe off the LF or the CR and leave it with only one, that would be helpful. Otherwise you get code that looks like

var a = "Hello!";

draw = function() {

Gigabyte-Giant:
Minor change; just attempts to resolve #333, by replacing `\r\n` (CRLF) line endings, with `\n` (LF) line endings.

The original regular expression (`/\r/g`), was replacing `\r`s with `\n`s.
CRLF line endings are a combination of `\r`s and `\n`s.
If the original replace was run against code with CRLF line endings, you'd end up with something like `\n\n`.

This PR replaces `\r\n` with `\n`, so we don't get an extra `\n`.

Gigabyte Giant (brynden.bielefeld@hotmail.com)

### DIFF
--- a/build/js/live-editor.editor_ace.js
+++ b/build/js/live-editor.editor_ace.js
@@ -392,7 +392,7 @@ window.AceEditor = Backbone.View.extend({
         if (_text != null) {
             this.editor.getSession().setValue(_text);
         } else {
-            return this.editor.getSession().getValue().replace(/\r/g, "\n");
+            return this.editor.getSession().getValue().replace(/\r\n/g, "\n");
         }
 
         return this;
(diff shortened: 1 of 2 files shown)


kevinbarabash:
That makes sense.  How did you test this?

Gigabyte-Giant:
I booted up my Windows 10 VM, and opened the demo, I then took a file that had CRLF line endings (provided by @spongejr), and pasted it into the editor, and reloaded the page...

Before the change, you got the extra newlines (as expressed in #333), but after the change, I was unable to get the extra newlines again.

kevinbarabash:
@Gigabyte-Giant sounds good.  Thanks for fixing this.

Gigabyte-Giant:
My pleasure. Thanks for reviewing.

Issue #400, opened 2015-08-25
calls to the .toString() on functions should output code that is the same as the user code


kevinb7:
This is used in programs such as the Button Creator, see https://www.khanacademy.org/computer-programming/button-generator/1847123742.

Right now the generated code includes references to __env__1293812739.

MikaalSky:
Im not sure overriding .toString would be a good thing to do.

kevinb7:
Yeah, but I'm not sure what else we could.  Any ideas?

Potato42:
It probably isn't a big deal to override .toString, and if we could do that, we could just add a .__toString or something, but how would it even work?  Sounds like sci-fi to me.

Gigabyte-Giant:
@kevinbarabash, @MikaalSky: We'd really only need to override `Function.prototype.toString`, correct?

kevinbarabash:
Yes that is correct.

kevinbarabash:
I have a plan for this which I think should work.  All `FunctionExpression` nodes in the AST should be rewritten to be IIFEs that return a `FunctionExpression` with a modified `.toString` method.  A function like `var foo = function() { console.log("foo"); }` should be rewritten as the following:
```
var foo = (function() { 
    var _foo = function() { console.log("foo"); }  // original function
    _foo.toString = function() { return __ka__toString(55, 73); }
    return _foo;
})();
```
Here, `__ka__toString` would be a function that's passed two parameters, the start and end of a character range in the original user code.  It would return the substring for that range.

lgto4:
Since https://github.com/Khan/live-editor/issues/449 is marked as a duplicate of this, the orignal function's text should include its orignal comments.

kevinbarabash:
@lgto4 it will.  I'm going to grab the function's source from the user code verbatim so it will contain the comments and all the original formatting.

PR #511, opened 2015-12-05
Make KAInfiniteLoopSetTimeout More Secure

### LINKED ISSUE #510: Users Can Effectively Disable Loop Proctection
Recently I found a program on the hot list that was (effectively) disabling the loop protector with this code.

```
var begin = function() {
    this[["KAInfiniteLoopSetTimeout"][0]](40000);
};
begin();
```

The author has since removed this code (as far as I can tell) after getting some flak for it.

I've fixed this issue in my fork of the live editor and should have a PR coming in shortly since this is a relatively simple fix.

JavascriptFTW:
This pull request fixes #510 by making KAInfiniteLoopProtect more secure.
.
## Changes

* Add a pseudo randomly generated ID to instances of `LoopProtector` using `Math.random()`. This technique should be secure enough as it's not likely that someones code will be able to guess 15+ digits.

* Adds the loop protectors ID as an argument to all calls to `KAInfiniteLoopTimeout` that are generated in the AST transform.

* Edits `LoopProtector.prototype._KAInfiniteLoopSetTimeout` to accept an extra argument which should be the ID of the loop protector. If the ID that it is passed does not match the loop protectors ID then an error will be thrown.

Merging this PR may need to be delayed until after HOC. I'm pretty sure that I didn't break anything but it _is_ messing with a rather important part of the live editor for beginners. I would rather not be responsible for ruining a beginners day.

Thank you for your time!

### DIFF
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -87416,6 +87416,10 @@ window.walkAST = function (node, path, visitors) {
  * @constructor
  */
 window.LoopProtector = function (callback, timeouts, reportLocation) {
+    // protectorKey is used to check if a call to KAInfiniteLoopSetTimeout is
+    // by our own code, or if the call is just a users code trying to mess with
+    // things.
+    this.protectorKey = Math.random();
     this.callback = callback || function () {};
     this.timeout = 200;
     this.branchStartTime = 0;
@@ -87511,9 +87515,18 @@ window.LoopProtector.prototype = {
         }
     },
 
-    _KAInfiniteLoopSetTimeout: function _KAInfiniteLoopSetTimeout(timeout) {
-        this.timeout = timeout;
-        this.branchStartTime = 0;
+    _KAInfiniteLoopSetTimeout: function _KAInfiniteLoopSetTimeout(timeout, protectorKey) {
+        // Make sure that the key we've been provided matches the one on this
+        // LoopProtector
+        if (protectorKey === this.protectorKey) {
+            this.timeout = timeout;
+            this.branchStartTime = 0;
+        } else {
+            // Let the user know that what they're trying to do is a bad idea
+            throw {
+                html: "Messing with the loop protector is dangerous!"
+            };
+        }
     },
 
     riskyStatements: ["DoWhileStatement", "WhileStatement", "ForStatement", "FunctionExpression", "FunctionDeclaration"],
@@ -87551,7 +87564,7 @@ window.LoopProtector.prototype = {
             // the timeout to mainTimeout just to reset the value when the program
             // is re-run.
             if (this.mainTimeout) {
-                node.body.unshift(b.ExpressionStatement(b.CallExpression(b.Identifier("KAInfiniteLoopSetTimeout"), [b.Literal(this.mainTimeout)])));
+                node.body.unshift(b.ExpressionStatement(b.CallExpression(b.Identifier("KAInfiniteLoopSetTimeout"), [b.Literal(this.mainTimeout), b.Literal(this.protectorKey)])));
             }
 
             // Any asynchronous calls such as mouseClicked() or calls to draw()
@@ -87559,7 +87572,7 @@ window.LoopProtector.prototype = {
             // responsive as the app is running so we call KAInfiniteLoopSetTimeout
             // at the end of main and set timeout to be asyncTimeout
             if (this.asyncTimeout) {
-                node.body.push(b.ExpressionStatement(b.CallExpression(b.Identifier("KAInfiniteLoopSetTimeout"), [b.Literal(this.asyncTimeout)])));
+                node.body.push(b.ExpressionStatement(b.CallExpression(b.Identifier("KAInfiniteLoopSetTimeout"), [b.Literal(this.asyncTimeout), b.Literal(this.protectorKey)])));
             }
         }
     },
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -87524,7 +87524,7 @@ window.LoopProtector.prototype = {
         } else {
             // Let the user know that what they're trying to do is a bad idea
             throw {
-                html: "Messing with the loop protector is dangerous!"
+                html: i18n._("Messing with the loop protector is dangerous!")
             };
         }
     },
--- a/js/output/shared/loop-protect.js
+++ b/js/output/shared/loop-protect.js
@@ -115,7 +115,7 @@ window.LoopProtector.prototype = {
         } else {
             // Let the user know that what they're trying to do is a bad idea
             throw {
-                html: "Messing with the loop protector is dangerous!"
+                html: i18n._("Messing with the loop protector is dangerous!")
             };
         }
     },
--- a/js/output/shared/loop-protect.js
+++ b/js/output/shared/loop-protect.js
@@ -11,6 +11,10 @@
  * @constructor
  */
 window.LoopProtector = function(callback, timeouts, reportLocation) {
+    // protectorKey is used to check if a call to KAInfiniteLoopSetTimeout is
+    // by our own code, or if the call is just a users code trying to mess with
+    // things.
+    this.protectorKey = Math.random();
     this.callback = callback || function () { };
     this.timeout = 200;
     this.branchStartTime = 0;
@@ -102,9 +106,18 @@ window.LoopProtector.prototype = {
         }
     },
 
-    _KAInfiniteLoopSetTimeout: function (timeout) {
-        this.timeout = timeout;
-        this.branchStartTime = 0;
+    _KAInfiniteLoopSetTimeout: function (timeout, protectorKey) {
+        // Make sure that the key we've been provided matches the one on this
+        // LoopProtector
+        if (protectorKey === this.protectorKey) {
+            this.timeout = timeout;
+            this.branchStartTime = 0;
+        } else {
+            // Let the user know that what they're trying to do is a bad idea
+            throw {
+                html: "Messing with the loop protector is dangerous!"
+            };
+        }
     },
 
     riskyStatements: [
@@ -173,7 +186,7 @@ window.LoopProtector.prototype = {
                 node.body.unshift(
                     b.ExpressionStatement(b.CallExpression(
                         b.Identifier("KAInfiniteLoopSetTimeout"),
-                        [b.Literal(this.mainTimeout)]
+                        [b.Literal(this.mainTimeout), b.Literal(this.protectorKey)]
                     ))
                 );
             }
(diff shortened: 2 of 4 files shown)


kevinbarabash:
@JavascriptFTW thanks for the pull request.  Can you test it with some of the programs in demos/simple/scratchpads.html?  In particular the tower hanoi scratchpads.

JavascriptFTW:
@kevinbarabash I tested the tower of hanoi scratchpads with a solution I found on Google and loop protect worked fine, and still caught things. I also tested with a few others (for example the primality test) and it also worked. I'm not entirely sure if this is what you were asking for though so sorry if it wasn't.

kevinbarabash:
@JavascriptFTW that's perfect.  Thanks.

thomasmorelli:
@JavascriptFTW @kevinbarabash I don't know if this works any longer, but I just saw [this program](https://www.khanacademy.org/computer-programming/infinite-slide/6611264526745600) that is overriding the definition of ```KAInfiniteLoopProtect``` so that it doesn't do anything. Please correct me if I'm wrong and it has already been fixed!

JavascriptFTW:
@thomasmorelli That program seems fine as if you `console.log(KAInfiniteLoopProtect);` after the user has supposedly no-opped it you still get the expected result.

thomasmorelli:
@JavascriptFTW I get ```bound () { [native code] }``` when I ```console.log(KAInfiniteLoopProtect);```- Is this expected? I'm sorry I don't really know much about this sort of thing

build/js/live-editor.output_pjs_deps.js
```diff
@@ -87523,9 +87527,18 @@ window.LoopProtector.prototype = {
...
+    _KAInfiniteLoopSetTimeout: function _KAInfiniteLoopSetTimeout(timeout, protectorKey) {
+        // Make sure that the key we've been provided matches the one on this
+        // LoopProtector
+        if (protectorKey === this.protectorKey) {
+            this.timeout = timeout;
+            this.branchStartTime = 0;
+        } else {
+            // Let the user know that what they're trying to do is a bad idea
+            throw {
+                html: "Messing with the loop protector is dangerous!"
```
kevinbarabash: I think this might conflict with changes from https://github.com/Khan/live-editor/commit/57130931aa8ded361a081b7cfc803c3bf4d317e0.  Could you rebase and then run the async tests to see if there's an issue?

js/output/shared/loop-protect.js
```diff
@@ -116,9 +120,18 @@ window.LoopProtector.prototype = {
...
+    _KAInfiniteLoopSetTimeout: function (timeout, protectorKey) {
+        // Make sure that the key we've been provided matches the one on this
+        // LoopProtector
+        if (protectorKey === this.protectorKey) {
+            this.timeout = timeout;
+            this.branchStartTime = 0;
+        } else {
+            // Let the user know that what they're trying to do is a bad idea
+            throw {
+                html: "Messing with the loop protector is dangerous!"
```
kevinbarabash: I think this might conflict with changes from https://github.com/Khan/live-editor/commit/57130931aa8ded361a081b7cfc803c3bf4d317e0.  Could you rebase and then run the async tests to see if there's an issue?  Also, this message should be wrapped in call to `i18n._`, e.g. `i18n._("Messing with the loop protector is dangerous!")`.

kevinbarabash:
@JavascriptFTW sorry for taking so long to review this.  The plan is to replace the infinite loop protector in the new year with something else, but since it's going to take some time to implement and more time to work out all of the bugs it's good to have this in place.  We'll want to test the new system with only a small portion of our users, during that time the rest of the users will continue using the loop protector.

JavascriptFTW:
@kevinbarabash It's okay. The holidays and all.

I've rebased, and wrapped the message in an i18n call as requested. I ran the tests in async_test.js on my machine and there were 3 failures. Travis CI had no problems when it tested though so I'm thinking I just did something wrong by accident. If this warrants concern and needs to be fixed just let me know and I'll get to it.

kevinbarabash:
The async_test.js tests don't get run on travis-ci.  The reason is that some of the tests throw errors that can't be caught and thus the tests fail.  I'm pretty sure that some of failures are pre-existing from changes in https://github.com/Khan/live-editor/commit/57130931aa8ded361a081b7cfc803c3bf4d317e0.  I'll check out your changes tomorrow.

DozenalFTW:
I still dislike it when AOAS WNW (all of a sudden, with no warning) a false error pops up, with no way to say, "**No, I'm fine. Continue with the program, please**."
That kind of feature would be really nice. Right now, the best way around it is to "back up" the entire state of the game every 3 seconds or so and then "rewind" if an infinite loop happens. Which is tedious, since you have to place your custom detector in every function.

I especially notice that the error pops up more when `focused` is false than when it's true. For example, once I was playing a game with really good graphics and the *SECOND* I clicked Spin-off, the game froze. So I think the loop-catch timer should be 2 seconds when the program is focused and 10 if it's not.

Also, you can likely ascertain the KAInfiniteLoopProtect random 15-digit number by creating a dummy function, calling `.toString()`, and locating `/[0-9]{15}/`. I'm sure that by studying the injections I can find the 15 digits.

MikaalSky:
@DozenalFTW I'm pretty sure it could be made near impossible to get the code, and I'm pretty sure it will be nearly impossible to get it. I do get your point with the continue, but saying "Oh i'll just circumvent the new security now" isn't a very.. +rep thing to say.

kevinbarabash:
It believe it would be impossible to get the transformed code if we fixed https://github.com/Khan/live-editor/issues/400.  Of course someone could download the source for live-editor and compile their code that way, but having a random number would make that a fruitless endeavour.

kevinbarabash:
@DozenalFTW the plan is to eventually get rid of the loop protector and implement the "No, I'm fine. Continue with the program, please." behaviour.  Unfortunately, this is really difficult to do because JavaScript is single thread so it doesn't support stopping execution halfway through a function except by throwing an exception.  Once an exception is thrown, there's no way to resume execution where you left off.  It's not impossible though.

JavascriptFTW:
Should I close this PR? This issue is not vital and it seems to be low priority.

kevinbarabash:
@JavascriptFTW sorry for waffling on this.  Let's close this for now.

EricBalingit:
For what it's worth, regarding #400, having my code munged any more than absolutely necessary, for any reason, makes me cringe.  Regardless, I prefer to see what the live editor has done to my code.

Nearly 100% of people before Khan Academy did not learn programming with a live editor and it seems to have worked out for them.  To me, the browser freezing tells me I have a broken loop and crashing tells me I probably have a memory leak.

My humble opinion is that "No, I'm fine..." is more obtrusive/less instructive in a way than simply interrupting the program - which boils down to requiring changes to be made in order for the intended result to happen ( which is no interruptions at all ).  At what point should the learner start relying on a long running construct?  At what point should they stop relying on that construct and make an effort to make the code more efficient?

To that end, the current loop protection is enough to keep the learner engaged and prevent the page from freezing - in other words, it is enough.  Going further than that, to me, seems rather over-the-top.  I do not see what the instructive value of such a thing would be.  And I would even venture to remind the educators here that yes, the sandbox is a fun and complex object, an intellectual curiosity even, but it's purpose, primarily, is to educate beginners, not to educate advanced engineers, even though that may happen as a consequence of its existence. :-)  So the question is - are ( name your proposed ) "improvements" for curiosity's sake or for the sake of making a better educational product?

And further, if it can be hacked on a page, without affecting the site, then why not?  Is that not part of learning like any other approach to software?  Do we not need creative thinkers and security experts?  This organization does have policies and procedures for handling issues of defacement of the site, and granted those should be exercised to the fullest and people should learn and understand what is acceptable behavior, and obviously, if an account holder is using the site inappropriately in a serious manner, then of course those issues must be handled effectively and appropriately - on an individual basis ( or am I wrong that KA does purport to deliver the custom educational experience ? ).  I feel that generally there is tolerance among Khan Academy moderators in such situations where code and its implied intent/behavior of its author are not threatening.

I have never been subject to warnings or other direct communications from moderators, possibly because I have a fair sense of what is appropriate behavior within the site, however, locking everything down and hiding the output ( internal/munged mechanism ) of the users code is sending the opposite message of the kind of behavior that is expected.  Recall that knowledge, and particularly knowledge of information technology, put bluntly, hacking code, is a form of power.  Overtly taking that power away may seem prudent, but it is a basis for mistrust and an invitation for mischief.  It is telling learners that they are not trusted, that they are suspect and that everything must be done to sterilize the environment in which they are invited to grow and learn.  I do not think that is the kind of message that Khan Academy wants to send to its students.

Maybe the cookie cutter effect of the education system has simply caused more damage than we all realize or are willing/able to admit.

DozenalFTW:
@EricBalingit I disagree with one part. The issue is NOT when a legitimately broken loop would otherwise freeze the tab. The issue is all the **false** alarms that were cast. I have many times done nothing but read and interact with the Tips and Thanks, and the program was stopped.
The "no I am fine let us continue" is for _those_ times, when it was just having a random burst of slowness for reasons that cannot be stopped. Or when you navigate away from the program and it (rightfully) uses less resources on the program, but the infinite loop detector thinks that is a problem.

I think five seconds is a good compromise in the meantime. A real infinite loop would terminate quickly and someone has time to do things if it is not legitimate. 500 milliseconds does not allow that freedom.

But I agree with the rest of your points.

EricBalingit:
@DozenalFTW So does making an edit to the code then trigger a closing of the "no I'm fine prompt" if it happens to be open at that time? All I'm really getting at is the current complexity of the software and the number of people working on it doesn't seem to permit too many bells and whistles. Maybe that will change eventually. I've looked at the code. I've looked over a couple of pull requests. I wouldn't want to spend my time making much of anything for the sandbox let alone trying to fix much of it. Just saying. :-)

kevinbarabash:
There are a number of competing concerns.  I think it's important for users to be able to peak behind the curtain, but our version of JavaScript should be as close to real JavaScript as possible while providing protection from infinite loops.

With the 500ms timeout, I was worried about animations where every frame takes multiple seconds.  This would result in a browser that's barely usable.  A possible improvement would be to have different timeouts based on whether the long running code is called from the `draw` loop or is in response to a one-off event like `mousePressed` or `mouseReleased`.

This project is in maintenance mode right now so there won't be any big changes for a while.

DozenalFTW:
Another issue is `frameRate`. Sometimes I have done really intense animations that take 100ms to complete normally and then I set like `frameRate(6)`.

I wonder if there is an easy way to have the program itself (i.e. user code, not behind-the-scenes code) detect problems like this before they happen. Some sort of `millis()` counter mixed with if statements after every possible point of failure. It could save the important game state into `Program` so the user can restart, or something.

EricBalingit:
@DozenalFTW Not sure what you mean, can you provide an example and then *comment on one of my programs*?  We should try to keep these discussions about sandbox code, not user code. :-)

PR #560, opened 2016-03-07
Change color() and getImage() auto-complete to not insert ; as often. Fixes #262.

### LINKED ISSUE #262: Autocomplete of color() inside array produces unnecessary semi-colon
For example:
```
var bla = [
  color: color(255, 0, 0);
];
```
We could decide to never put the semi-colon for that autocomplete. Or be smarter via regex/ast. 
Or decide that students can figure it out, which I'm sure they do. :-)

jeresig:
I've changed the behavior so that the following now happens for both color and getImage:

No semicolon:

```
color(
var a = [color(
var a = [
    color(
foo(color(
foo(
    color(
if (a === color(
```

Semicolon:
```
var a = color(
```

### DIFF
--- a/build/js/live-editor.tooltips.js
+++ b/build/js/live-editor.tooltips.js
@@ -1649,6 +1649,11 @@ window.TooltipBase = Backbone.View.extend({
         return parenStack.length > 0;
     },
 
+    // Returns true if we're after an `=` assignment
+    isAfterAssignment: function isAfterAssignment(text) {
+        return /(?:^|[^!=])=\s*$/.test(text);
+    },
+
     // Returns true if we're inside a string
     isWithinString: function isWithinString(text) {
         var withinString = false;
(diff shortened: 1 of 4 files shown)


kevinbarabash:
Yay!  Thanks for fixing this long standing issue.  I'll review this when I get into the office.

kevinbarabash:
Looks good to me.  I wish that we could test some of these behaviors better.  Ideally it would be nice if we had a separate library that we could throw a block of text at along with a cursor position and it would return what would be suggested or autocompleted.  That's going on the wish list.

pamelafox:
Hm, I'm writing background() now and not seeing a semi-colon inserted after. I don't think this PR was intended to change the way background() works, right?

kevinbarabash:
This also applies to `stroke()` and `fill()`.  This shouldn't be too hard to fix.

PR #571, opened 2016-04-07
Changes to structure()

SlySherZ:
Makes patterns indistinguishable from structures, which is helpful if we want to have base patterns that include constraints, and build on top of those.
I considered multiple options to check for in the if clause and decided this is the best one because patterns can have many different types.


kevinbarabash:
I'm hesitant to merge this because I'm not sure what effect this might have on existing test code we have in production.  I'm going to hold off on merging this until we have a way to automatically run the regression tests for the challenges.

kevinbarabash:
@SlySherZ I don't have time to properly review this.

Issue #566, opened 2016-03-23
fill() autocomplete doesn't add a semi-colon anymore


pamelafox:
There as a recent PR that (I believe) was intended to just change autocomplete behavior for color() and getImage(). However, I think it affected fill(), as it no longer adds a semi-colon. I've found this frustrating over the last few days, as it means I have to get my cursor back outside of the colors in order to add a semi-colon. With fill() and stroke(), I always want the semi-colon.

pamelafox:
@kevinbarabash Can you file this internally? I'd suggest reverting john's recent change as a quick fix. Thanks!

kevinbarabash:
@pamelafox I'll fix this on the weekend.  I stubbed my toe on it too.

MikaalSky:
Please fix asap, it's very annoying when just writing code. :)

kevinbarabash:
I'll do a deploy sometime this week.

MatthiasSaihttam:
This issue really should be closed, as the bug is fixed.

kevinbarabash:
@MatthiasSaihttam thanks.  Someone else must've deployed live-editor to prod.

PR #573, opened 2016-04-10
Add semicolons unconditionally for fill/stroke function calls

kevinbarabash:
Since fill/stroke commands shouldn't appear as args to other
function calls it's okay to append a semicolon.

Test Plan:
- load http://localhost:63343/live-editor/demos/simple/index.html?new_error_experience=yes
- type fill( and see the semicolon being added at the end of the autocomplete
- do the same for stroke(
- type color( and see the semicolon not being added
- do the same for getImage(
- assign color( and getImage( to variables and verify that semicolons are being added

### DIFF
--- a/js/ui/tooltips/color-picker.js
+++ b/js/ui/tooltips/color-picker.js
@@ -101,11 +101,18 @@ TooltipEngine.classes.colorPicker = TooltipBase.extend({
         };
         this.aceLocation.tooltipCursor = this.aceLocation.start + this.aceLocation.length + this.closing.length;
 
-        if (event.source && event.source.action === "insertText" && 
+        var name = event.line.substring(functionStart, paramsStart - 1);
+        var addSemicolon =
+            this.isAfterAssignment(event.pre.slice(0, functionStart));
+        if (['fill', 'stroke'].includes(name)) {
+            addSemicolon = true;
+        }
+
+        if (event.source && event.source.action === "insertText" &&
             event.source.text.length === 1 && this.parent.options.type === "ace_pjs" && this.autofill) {
             // Auto-close
             if (body.length === 0 && this.closing.length === 0) {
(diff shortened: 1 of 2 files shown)


kevinbarabash:
Fixes https://github.com/Khan/live-editor/issues/566.

js/ui/tooltips/color-picker.js
```diff
@@ -101,11 +101,18 @@ TooltipEngine.classes.colorPicker = TooltipBase.extend({
         };
         this.aceLocation.tooltipCursor = this.aceLocation.start + this.aceLocation.length + this.closing.length;
 
-        if (event.source && event.source.action === "insertText" && 
+        var name = event.line.substring(functionStart, paramsStart - 1);
+        var addSemicolon =
+            this.isAfterAssignment(event.pre.slice(0, functionStart));
+        if (['fill', 'stroke'].includes(name)) {
```
jeresig: Do we have .includes support polyfilled? May be better to just use indexOf, just in case.

jeresig:
@kevinbarabash Thanks for fixing this! One possible bug fix inline.

js/ui/tooltips/color-picker.js
kevinbarabash: We include `browser-polyfill.min.js` in `live-editor.shared.js`.  Also, `includes` is already in use in other places of the code.

jeresig: Ok, good to know! Fine by me then :)

MatthiasSaihttam:
Right now semicolons are not being added after a `background()` command. `background` should really be included with `stoke` and `fill` in that array. I'm very new to Github, how could I help get this fixed?

SharkFinPro:
I have noticed this issue too, you could [directly tell the devs...](https://gitter.im/Khan/live-editor)

PR #580, opened 2016-04-30
Update structuredjs

MrMagma:
This PR updates the structuredjs submodule to the latest commit.


kevinbarabash:
Hmm... looks like it's been a while since we've updated and now it's causing test failures.  @JavascriptFTW I think your changes are safe, but some of the earlier changes are not.  I need to think about how we can get your changes in without pulling in the other changes.

kevinbarabash:
@MrMagma sorry I don't have time to review this PR.

PR #581, opened 2016-05-04
Add 'stopSound' command

lafisrap:
With a musican friend I plan to write some small educational music apps/games that students can modify. For that we need a stopSound command.

I tested this one with 5 parallel sounds.

Format: stopSound( sound, fadeOutTime )

sound: A sound retrieved by getSound()
fadeOutTime: Time in seconds the sound is fading out

### DIFF
--- a/js/output/pjs/pjs-code-injector.js
+++ b/js/output/pjs/pjs-code-injector.js
@@ -156,12 +156,45 @@ class PJSCodeInjector {
 
             playSound: (sound) => {
                 if (sound && sound.audio && sound.audio.play) {
+					if( sound.audio.fadeOutInterval ) {
+						sound.audio.fadeOutStopInterval();                                    
+					}
+
                     sound.audio.currentTime = 0;
                     sound.audio.play();
                 } else {
                     throw {message: i18n._("No sound file provided.")};
                 }
             },
+
+			stopSound: (sound, fadeOut) => {
+				if (sound && sound.audio && sound.audio.volume) {
+					if( !fadeOut || fadeOut < 0.020 ) {
+						sound.audio.pause();
+					} else {
+						var volume = sound.audio.volume,
+							orgVolume = volume;
+
+						sound.audio.fadeOutInterval = setInterval( function() {
+							volume -= orgVolume / 20;
+							if( volume > 0 ) {
+								sound.audio.volume = volume;
+							} else {
+								sound.audio.fadeOutStopInterval();                                    
+							}
+						}, fadeOut * 1000 / 20 );
+
+						sound.audio.fadeOutStopInterval = function() {
+							clearInterval( sound.audio.fadeOutInterval );
+							sound.audio.fadeOutInterval = null;
+							sound.audio.pause();
+							sound.audio.volume = orgVolume;
+						};
+					}
+				} else {
+                    throw {message: i18n._("No sound file provided.")};
+				}
+			},
 
             // Basic console logging
             debug: (...args) => {
--- a/tests/output/pjs/output_test.js
+++ b/tests/output/pjs/output_test.js
@@ -52,6 +52,21 @@ describe("Scratchpad Output Exec", function() {
         test("getSound with computed string", function() {
             getSound('rpg/' + 'giant-no');
         });
+
+        test("stopSound with different fadeOuts", function() {
+            var s1 = getSound("rpg/giant-no"),
+                s2 = getSound("rpg/giant-no"),
+                s3 = getSound("rpg/giant-no");
+
+            playSound(s1);
+            stopSound(s1, 0);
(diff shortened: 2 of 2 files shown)


kevinbarabash:
@lafisrap thanks for the pull request.  Can you add some automated tests for this?

lafisrap:
Just did ;-)

js/output/pjs/pjs-code-injector.js
```diff
@@ -156,13 +156,46 @@ class PJSCodeInjector {
...
                 }
             },
 
+			stopSound: (sound, fadeOut) => {
+				if (sound && sound.audio && sound.audio.volume) {
+					if( !fadeOut || fadeOut < 0.020 ) {
+						sound.audio.pause();
+					} else {
+						var volume = sound.audio.volume,
+							orgVolume = volume;
```
kevinbarabash: Can you make these two separate `var` declarations?  

Also, what if `stopSound` gets called during fadeout?  It might be safer to set `orgVolume` to whatever the default volume for `<audio>` tags seeing as there's no processing API to set the volume.

tests/output/pjs/output_test.js
```diff
@@ -249,12 +264,18 @@ describe("Scratchpad Output Exec", function() {
...
     });
 
-    failingTest("getSound with no string", function () {
+    failingTest("playSound with no string", function () {
         var s = getSound("");
 
         playSound(s);
     });
 
+    failingTest("stopSound with no string", function () {
```
kevinbarabash: What happens if you stop a sound that was never playing or stop a sound multiple times?

kevinbarabash:
@lafisrap sorry for taking so long to get back to you.  I left some comments in the code.  Could you switch from hard tabs to 4 space soft tabs?  Also, could you check in the built files for these changes?

js/output/pjs/pjs-code-injector.js
lafisrap: When stopSound() is called during fadeout I (now ;-) stop the running interval when it is called without fade option and just return and do nothing when it is called with fade option. It would be possible to let it fade at the second (or consecutive calls) from whereever the volume is by then, but this is a higher complexity level and I am not sure if there are use cases for it.

What do you mean with default volume of the audio tag? I only know the volume property and found nothing in stackoverflow.

lafisrap:
Hi Kevin, I have some issues with my fork of live-editor. I can't build it with the error message "this.merge is not a function" I read about this here in the issues and followed the recommendations (renaming handlebars) but it doesn't work. It does with a direct clone of live-editor from khan. So I made the changes there now and tested them. Sees to work fine. But I don't know how to get a proper github pull request going. I could send you patches (like in former times ;-) )

lafisrap:
I also tried to fork live-editor again, with the same outcome. That's why this pull request now comes from an unknown repository ;-) You have an idea why this merge error occurs? gulp build seems to work fine.

kevinbarabash:
:(
Can you send me a link to your branch so that I can check it out and see what's going on?

lafisrap:
I pushed all changes to my repo: https://github.com/lafisrap/live-editor

lafisrap:
What I did:
- I forked khan/live-editor with github.com fork button. 
- I cloned it to my computer
- Then: git submodule update --init --recursive
- sudo npm install
- bower install
- gulp build
First commit
- Made the changes in the source files
- gulp build
Second commit

kevinbarabash:
Thanks for the detailed set of instructions.  This will be really helpful in reproducing the issue.

lafisrap:
If it would help if I do the setup again and send you all the output, than I would do so ;-) There were some warnings ...

SharkFinPro:
Is this gonna ever be done?

js/output/pjs/pjs-code-injector.js
kevinbarabash: I assumed that you were setting `orgVolume = volume` so that it can be reset at a later time when `fadeOutStopInterval` is called.  The problem I see with this code is that if `stopSound` is called during a fade out `orgVolume` could be less than one might expect.  Sorry, I'm probably missing something obvious, but I don't see where the interval is cancel when calling `stopSound()`.

kevinbarabash:
@lafisrap I checked the repo and noticed https://github.com/lafisrap/live-editor/commit/31212f77774b8c580ab91cad9b1c23387bc073ad.  I'm confused by this pull request b/c those changes should show up.  Also, the branch that you're merging from at the top says `unknown repository`.

lafisrap:
Hej Kevin, sorry that it took so long until I found my way back here. I deleted the original repo for some stupid reason, so I set it up again, added a tooltip, and tested it with my live-editor setup. It should work.

PR #584, opened 2016-05-22
Add Duskpin

jcstudios:


MikaalSky:
Why just duskpin

jcstudios:
Because it's my favorite.

kevinbarabash:
@jcstudios sorry for taking so long.  Could you change this to be a vertically formatted array so that it's easier to see additions in the future.  Also, could you run `gulp build` and include the build files?

kevinbarabash:
whoops... didn't mean to close this.

MatthiasSaihttam:
I was think about this, and wondered if we should be adding the hidden avatars. Previously (I believe) avatars haven't been (formally) added to the live editor until after they've added to the site and have been made obtainable. If we do go ahead and add Duskpin, shouldn't we also add the other "hidden" avatars (Primosaur and Starky)? If you'd like them to be added I'd be willing to make the code changes.

kevinbarabash:
@MatthiasSaihttam we can add those other avatars as well.

kevinbarabash:
@jcstudios friendly ping.

Gigabyte-Giant:
This should be closed, as-per https://github.com/Khan/live-editor/commit/e20cfbbea7097d595391281c935549c3c215d3f3.

rileyjshaw:
Oops, I didn't see this - sorry for duplicating your work @jcstudios!

Duskpin (and friends) will be on the site soon 😄

Issue #551, opened 2016-02-05
parseInt() returns 0 instead of NaN if given invalid input


darrylyeo:
The following examples should print `NaN`, not `0`:
```println(parseInt("", 10));
println(parseInt("Hello World", 10));
println(parseInt("42", 2));```

kevinbarabash:
processing-js re-implements `parseInt` to follow different semantics.  I believe we should be use the JavaScript semantics.  This should be less confusing for our users.

DozenalFTW:
And while we're at it, don't _require_ use of a second parameter.

kevinbarabash:
The second parameter is a best practice so it should at the very least be a warning.  Without if the number starts with a `0` then it parses the string as an octal number.

kevinbarabash:
I guess we could warn about numbers starting with `0`.

PR #600, opened 2016-08-14
Javascript's parseInt and new Images

MatthiasSaihttam:
This includes the changes suggested in #584, adding the hidden avatars to the image picker, making that PR obsolete.
It also changes `parseInt` to use native Javascript instead of Processing's code, fixing #551. 

Edit: This appears to be including changes from my last PR, that have already been merged. It's only the last three commits that need to be included, but in theory it wouldn't hurt to re-do the changes already made.


kevinbarabash:
This is very weird.  It looks like github is confused.  Could you create separate pull requests for the images additions and the `parseInt` change?

MatthiasSaihttam:
Sure, it might take me a while, but I'll see what I can do.

Gigabyte-Giant:
Just a quick word of advice @MatthiasSaihttam:

Whenever you get ready to fix a bug, or add a feature, **[create a new branch](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging)** in your fork of the live-editor repository.
Doing so, makes it much easier to submit pull requests, as you'll just request to merge `MatthiasSaihttam:yourFeature` into `Khan:master`.

MatthiasSaihttam:
Yea, I'm going to start doing that. I figured it would make things more complicated having to worry about different branches, but it's probably worth it. Thanks for the advice.

kevinbarabash:
Doing PRs from branches adds a little overhead, but they're super helpful if you want to update your master to to match `khan/live-editor`'s `master`.

Gigabyte-Giant:
Exactly what @kevinbarabash said. It takes a little bit more effort, but, in my opinion, it's worth it in the long run.

They can definitely also help prevent problems like was experienced with this PR (multiple commits that've already been merged).

Issue #601, opened 2016-08-21
Alternative constructors don't work


Potato42:
An idiom I sometimes use in JS (that somewhat mimics Python's `classmethods`) is 
```javascript
Constructor.fromSomething = function(something) {
    return new this(...);
};
```

I am pretty sure it used to work on KA, but now I get cryptic errors for simply creating such an alternative constructor.  See [this example program](https://www.khanacademy.org/computer-programming/watfromkhanacademyjs/6294900650672128).

The error seems to be coming from the code rewriting or something as you can't catch it.

kevinbarabash:
I don't think this is a pattern we want to encourage.  The same effect can be accomplished with:
```
Constructor.fromSomething = function(something) {
    return new Constructor(something);
}
```

Potato42:
First off, it doesn't really matter if it is a good pattern, it is a bug that it doesn't work and has a cryptic error message.  I wouldn't be surprised if there are other related bugs.

Also, your way doesn't work with subclasses, while my way works as long as you copy properties attached to the constructor over.

Any ideas where this problem is coming from?

kevinbarabash:
> Also, your way doesn't work with subclasses, while my way works as long as you copy properties attached to the constructor over.

Good point.

> Any ideas where this problem is coming from?

My guess is the `applyInstance` doesn't handle this case correctly.  We transform all `NewExpressions` into calls to `PJSOutput.applyInstance`.  See https://github.com/Khan/live-editor/blob/master/js/output/pjs/pjs-output.js#L559-L588.

Potato42:
I did a little digging, and I think I found [the culprit](https://github.com/Khan/live-editor/blob/0c7ef693aa4ba655e9922269c80a40e0a4c0980e/js/output/pjs/pjs-ast-transforms.js#L297-L303).
That function returns undefined when `node` isn't an `Identifier` or `MemberExpression`.  That undefined gets passed into `Literal`, which becomes a property called `value`, which is why you get an `e.value is undefined`.

**_I started to ramble on after here, so if you want the TL;DR it's "try using `escodegen.generate` instead of `memberExpressionToString`"_**

In our case, we're dealing with a `ThisExpression`, but there are other things you can use too that have the same problem.  For example:

```javascript
var getMyCoolConstructor = function() {
    return function(potato) {
        this.potato = potato;
    };
};

new (getMyCoolConstructor())(42);
```
and
```javascript
// jshint supernew: true
new (function(potato) {
[...]

PR #597, opened 2016-07-22
Fixed colorMode issue: Too many arguments passed!

aparms:
The `colorMode` issue is in khan/live-editor#576.

I only edited autosuggest-data.js The other files were changed during the build process.

PR #602, opened 2016-08-25
Used escodegen instead of memberExpressionToString

Potato42:
Removed the function `memberExpressionToString` and replaced it with
`escodegen.generate`.  The reason for this change is that escodegen
does everything `memberExpressionToString` does, but handles all node
types, which is important for some code.  See #601.

### DIFF
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -87835,23 +87835,11 @@ ASTTransforms.findResources = function (resources) {
     };
 };
 
-/**
- * Converts a MemberExpression to a String, e.g. given an AST for a.b.c will
- * return the string "a.b.c".
- */
-var memberExpressionToString = function memberExpressionToString(node) {
-    if (node.type === "Identifier") {
-        return node.name;
-    } else if (node.type === "MemberExpression") {
-        return memberExpressionToString(node.object) + "." + node.property.name;
-    }
-};
-
 ASTTransforms.rewriteNewExpressions = function (envName) {
     return {
         leave: function leave(node, path) {
             if (node.type === "NewExpression") {
-                var _name = memberExpressionToString(node.callee);
+                var _name = escodegen.generate(node.callee);
 
                 return b.CallExpression(b.CallExpression(b.MemberExpression(b.MemberExpression(b.Identifier(envName), b.Identifier("PJSOutput")), b.Identifier("applyInstance")), [node.callee, b.Literal(_name)]), node.arguments);
             }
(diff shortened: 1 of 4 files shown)


Potato42:
I don't know why, but the building process changed some `x` vs `&times;` thing.  Might have been a previous commit or something?

kevinbarabash:
I had a look at the tests and it looks there's a failure even those travis says it's a pass.
```
The following errors occurred:
    Error: KA_INFINITE_LOOP
        file:///home/travis/build/Khan/live-editor/build/js/live-editor.output_webpage_deps.js:11955
        file:///home/travis/build/Khan/live-editor/build/js/live-editor.output_webpage_deps.js:11980
        file:///home/travis/build/Khan/live-editor/build/js/live-editor.core_deps.js:155
        :1
        file:///home/travis/build/Khan/live-editor/tests/output/webpage/index.html?tests=output_test.js:2
    ReferenceError: Can't find variable: bla
        file:///home/travis/build/Khan/live-editor/tests/output/webpage/index.html?tests=output_test.js:
```
Does this test pass for you locally?

Potato42:
I noticed those errors earlier, but I don't think that's my problem.  That has to do with the webpage environment, which my code shouldn't touch, and those errors [seem to be coming from](https://github.com/Khan/live-editor/blob/51c89d39f2055f5e8437e7a6f012d07fc2792631/tests/output/webpage/output_test.js#L268) within `failingTest`s, so based on [this](https://github.com/Khan/live-editor/blob/0c7ef693aa4ba655e9922269c80a40e0a4c0980e/tests/output/webpage/test_utils.js), it seems like the fact that errors are being raised is intentional -- `bla(x)` is pretty obviously supposed to raise an error.

As for why the test runner is showing the errors, I'm not really sure.  My best guess is that the text of the error you are testing against changed.  "Your javascript encountered a runtime error" is different from "Can't find variable" or something like that.  That doesn't explain how it passes -- if the errors are different it should fail I would think, but the exact same errors occur in #597's [travis build](https://travis-ci.org/Khan/live-editor/builds/148455091#L801) (and probably other builds too), so I think it is an issue with your tests, not an issue with my pull request.

kevinbarabash:
@Potato42 good point about it being the webpage environment.

kevinbarabash:
@Potato42 I tried out the change with a few different KA scratchpads and ran the tests locally.  It seems to work fine.  Thanks for the fix.  Can you sign Khan Academy's [Contributor License Agreement](https://khanacademy.org/r/cla) before I merge this?

Potato42:
Ok, I have signed it.

kevinbarabash:
@Potato42 awesome, thanks!

Issue #606, opened 2016-09-17
Urgent. Window unsealable in Safari 10 totally breaks live editor.


MatthiasSaihttam:
This is actually a pretty simple issue. In Safari 10, webkit [updated](https://trac.webkit.org/changeset/205404) the `window` object to be unsealable, instead throwing a TypeError. The problem is that the live-editor currently seals the window object ([here](https://github.com/Khan/live-editor/blob/0c7ef693aa4ba655e9922269c80a40e0a4c0980e/js/output/pjs/pjs-output.js#L113)), throwing a TypeError and preventing any other code from running. Right now I'm running Safari Technology Preview, and I just get the blue spinner forever when I try to use the live editor. This is more of a problem because Apple will ship macOS Serria (with Safari 10) on Tuesday the 20th, and this issue should ideally be fixed before then. I'll try to get a PR made within the next few hours, but my local repositry is pretty messed up, so I might fail.

Edit: `Object.freeze(window);`, what we do for Chrome, also doesn't work. Totally removing the check for Safari will work, as `Object.getOwnPropertyDescriptor(window).configurable` is false.

Edit 2: PR successfully created!

Potato42:
Obviously merge that pull request, but we should consider if freezing window actually is a good idea.  In the webpage environment, window is not frozen, but 
```javascript
window.location = "//example.com"
```
results in 
```Content Security Policy: The page's settings blocked the loading of a resource at https://example.com/ ("child-src https://*.kastatic.org https://www.khanacademy.org https://www.kasandbox.org").```

This is not the case for the PJS environment.  If you use Firefox and do 

```javascript
Object.constructor("this.location = '//example.com'")();
```

and it works!

Why is the HTML security ***better*** than the PJS security without even freezing window?  Can we just do whatever the HTML environment is doing in the PJS environment and get rid of bugs like this permanently?  Because at this point, freezing window pretty much only works in Chrome.

kevinbarabash:
> Why is the HTML security better than the PJS security without even freezing window? Can we just do whatever the HTML environment is doing in the PJS environment and get rid of bugs like this permanently? Because at this point, freezing window pretty much only works in Chrome.

@Potato42 good question.  I would like to eventually move the two environments closer together for this and other reasons.

PR #607, opened 2016-09-17
Fix live-editor in Safari 10; don't freeze the window

MatthiasSaihttam:
This fixes issue #606 by not trying to freeze the window in Safari. As a result, the live editor works in Safari 10 instead of throwing a TypeError.


kevinbarabash:
@MatthiasSaihttam thanks for the quick fix.

PR #608, opened 2016-09-20
Use javascript's parseInt.

MatthiasSaihttam:
Per @kevinbarabash's comment on #551, changes the `parseInt` function back to using the native Javascript code instead of what processing.js sets it to. Fixes issue #551.

### DIFF
--- a/build/js/live-editor.output_pjs.js
+++ b/build/js/live-editor.output_pjs.js
@@ -2649,6 +2649,8 @@ window.PJSOutput = Backbone.View.extend({
 
         this.processing = new Processing(canvas, function (instance) {
             instance.draw = _this.DUMMY;
+
+            instance.parseInt = parseInt;
         });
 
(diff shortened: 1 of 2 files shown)


kevinbarabash:
@MatthiasSaihttam looks good.  Can you add a test?

MatthiasSaihttam:
If you don't mind, I'm thinking I'll totally replace [this](https://github.com/Khan/live-editor/blob/0e16a482a2d471fc9ff7e39a7d7a8c0c7305fb0d/tests/output/pjs/output_test.js#L445) test with a few of my own. That test seems confusing, and I'm not sure what it's trying to do.

kevinbarabash:
> That test seems confusing, and I'm not sure what it's trying to do.

I'm not sure either.

kevinbarabash:
Closing b/c we're in maintenance mode. See README.md for details.

PR #618, opened 2016-10-31
Use rimraf in npm install script

MattMS:
Using [rimraf](https://www.npmjs.com/package/rimraf) makes `npm install` Windows-friendly.


kevinbarabash:
@MattMS thanks for making things better for Windows developers.

PR #620, opened 2016-11-14
Add build READMEs

MattMS:
I wrote some notes while going through the `build` folder and `gulpfile.js`.

Maybe these would be useful for other people when tracking where the `build` files come from?

Alternatively, these notes could be added to the main README.


kevinbarabash:
@MattMS putting it in the main README.md makes more sense.

kevinbarabash:
Closing b/c we're in maintenance mode.  See README.md for details.

PR #622, opened 2016-11-16
Specify npm scripts in README

MattMS:
Simplify the commands in the README by using the npm scripts instead.

They don't need to have Gulp installed globally since it is included as a dev dependency.


kevinbarabash:
@MattMS nice clean up.  Thanks.

PR #625, opened 2016-12-16
Added stopSound() function.

lafisrap:
Parameter:
sound:	    as returned by getSound()
fadeout:    time to fade out in seconds

Tests and tooltips are also updated.


lafisrap:
I would like to handle the second parameter `fade out time` as optional. Is there a way to have optional parameters that are handled correctly by the syntax checker?

MatthiasSaihttam:
If I understand your problem correctly, you just need to add `stopSound` and the ways to use it to [this object](https://github.com/Khan/live-editor/blob/acd599ba9e20f37cc149027a9cb84490db9591b7/js/output/pjs/babyhint.js#L73). That'll fix Error Buddy saying "`stopSound` takes two parameters, not one."

build/js/live-editor.output_pjs.js
```diff
@@ -194,13 +194,53 @@ var PJSCodeInjector = (function () {
...
+                stopSound: function stopSound(sound, fadeOut) {
+                    if (sound && sound.audio && sound.audio.volume) {
+
+                        if (sound.audio.fadeOutStopInterval) {
+                            clearInterval(sound.audio.fadeOutInterval);
+                            sound.audio.fadeOutInterval = null;
+                        }
+
+                        if (!fadeOut || fadeOut < 0.020) {
+                            if (sound.audio.fadeOutStopInterval) sound.audio.fadeOutStopInterval();else sound.audio.pause();
```
kevinbarabash: Can you put these on separate lines?

kevinbarabash: with braces

build/js/live-editor.output_pjs.js
```diff
@@ -194,13 +194,53 @@ var PJSCodeInjector = (function () {
...
+                        } else {
+                            if (sound.audio.fadeOutStopInterval) return;
```
kevinbarabash: Please add braces.

kevinbarabash:
@lafisrap thanks for picking this up again.  The logic looks good.  There are some lint violations, but we don't seem to have the linter set up to run.  Could you run `eslint src` and fix any issues there?  I'm going to check out your diff and play around with it.

lafisrap:
Hej Kevin,
I just checked eslint with js/output/pjs/pjs-code-injector.js and it results in 144 errors. I installed eslint with "npm install eslint -g" Is this what you meant? Or do I miss some parameters?

lafisrap:
@kevinbarabash Hi Kevin, I still need this hint how to proceed with eslint?

kevinbarabash:
@lafisrap live-editor is in maintenance mode now b/c of changing priorities so we're looking to limit the changes we make.

Issue #448, opened 2015-10-16
Yoko Oh Noes does not  know that the rect function has eight parameters


lgto4:
Yoko Oh Noes needs to get up to speed wrt the Processing.js rect function.

`var rectangle = rect;
rectangle(200, 100, 150, 200, 0, 25, 0, 40);`

MikaalSky:
True. Current workaround until this gets fixed to be more lenient:
rect(x,y,sX,sY,
r1,r2,r3,r4);

kevinbarabash:
This should be a pretty easy issue to fix.  A person should be able to add additional optional params to the entry for `rect` in autosuggest.js.

Gigabyte-Giant:
Just in case anyone is interested, and if I'm not terribly mistaken, the changes should go [here](https://github.com/Khan/live-editor/blob/master/js/ui/autosuggest-data.js#L21-L33).

alexristich:
Hmm, this one's a little trickier than it first appeared. Adding tl, tr, br and bl params is possible, but it doesn't make sense to just tack them on as is. The current `rect` has `radius` as the fifth optional param, so adding these onto the back either means (a) having `radius` along with tl, tr, br and bl and having a clever way to ignore the fifth param, or (b) replacing `radius` with these more specific options, which allows for more control but less simplicity.

A final idea would be to come up with a way to, when a 6th param is entered, trigger a switch to another version of `rect` which has 8 params instead of 5 (thereby replacing `radius` in the tooltip with the four alternate options), but this might be more trouble than it's worth at the moment.

Thoughts?

kevinbarabash:
@alexristich I think that we'd probably have to rework how we're storing params for overloaded functions.  Right now we only have the concept of optional params.  What I'd like to have is a list of function signatures that a user could cycle between when the autohint pops up.  I think that would address the issues that you've identified.

alexristich:
Funny you mention that. When I first used the editor I wondered myself whether something like that might be better. I've been thinking about it for a bit here and, aside from any UI decisions, have a few questions of how this might best be accomplished. 

Presumably it would need to include the following:
1) An update to Oh Noes error checking logic so that it searches all the functions with a specified name
2) A way to suppress duplicate function names in the autohint popup (when typing a function name)
[...]

PR #643, opened 2017-08-15
Fixed param count for `rect`

EthanLuisMcDonough:
Addresses part of #448.


kevinbarabash:
Hmm... the version of node that travis-ci is using is super old.  I'll fix that over the weekend and that should fix the travis build.

JettBurns14:
_4 months later_ 🤔

PR #655, opened 2018-06-21
Get live-editor tests running on ChromeHeadless

pamelafox:
Background: I was writing another small PR, saw that the tests weren't passing, struggled to get them reliably working in Phantom, so I tried porting to ChromeHeadless, and got them all working there.

Where I've run the tests: 
- in the browser (with the Chrome flag to allow access from files, to get the workers working)
- command-line
- TravisCI

High-level overview of changes:
- I replaced gulp-mocha-phantom with gulp-mocha-chrome (used by a non-enabled, non-iframe test)
- I rewrote gulp-mocha-runner to use the Chrome headless API, based on what I learnt from the gulp-mocha-chrome codebase (to run all the enabled tests, which all run inside an iframe)
- I enabled the SQL tests, as they work just fine in Chrome!
- I enabled the async tests, with 2 modifications to get them working.

Test modifications:
- LoopProtector on mouseClicked: The loop protector throws an exception, to halt execution on the page, but mocha catches that and counts it as a failure. The other LoopProtector tests work because the protection function is triggered immediately, and thus could be wrapped inside a try/catch. However, triggering it on a user event makes it hard to try/catch. After trying many things, I decided to simulate the mouseClick by running "mouseClick()" as a second code block. That doesn't quite test what the original did, but that's what I came up with.
- draw updating global variables: This test checked that the ellipse() function got re-called with an increased diameter. It was intermittently failing, and I saw that by adding a short timeout, I could get it to work reliably. I believe that's due to the need for Processing to execute the next frame.


lukekrikorian:
Ok, this is epic

AqilCont:
Whoa... Pamela is here

JettBurns14:
Whoa, awesome! 🙌

CosignCosine:
Activity on the live editor??? There is hope once more!

kevinbarabash:
> After trying many things, I decided to simulate the mouseClick by running "mouseClick()" as a second code block. That doesn't quite test what the original did, but that's what I came up with.

I'm glad you found a workaround.

> I saw that by adding a short timeout, I could get it to work reliably. I believe that's due to the need for Processing to execute the next frame.

That makes sense.  Cool.

gulpfile.js
```diff
@@ -147,7 +147,7 @@ var runTest = function(fileName) {
         server.listen(11537);
 
         // We then run the Mocha tests in a headless PhantomJS
-        var stream = mochaPhantomJS();
+        var stream = mochaChrome();
```
kevinbarabash: It looks like a drop-in replacement.  Nice.

tests/output/pjs/async_test.js
```diff
@@ -218,15 +218,19 @@ describe("LoopProtector", function() {
             };
         });
 
+        var code2 = getCodeFromOptions(function() {
+            mouseClicked();
```
kevinbarabash: There isn't really an advantage to using `getCodeFromOptions` here, but it doesn't hurt any either either.

tests/output/pjs/iframe.html
```diff
@@ -12,19 +12,6 @@
         <div id="output-area"></div>
         <script src="../../../build/js/live-editor.core_deps.js"></script>
         <script>
-        // Works around this issue in PhantomJS:
-        // https://github.com/ariya/phantomjs/pull/12307
-        // Remove once PhantomJS pushes out a new release that fixes this!
```
kevinbarabash: Yay!

tests/output/pjs/test_utils.js
```diff
@@ -254,19 +254,8 @@ var getCodeFromOptions = function(code) {
...
-    ev.initMouseEvent(
-        "mouseup",
-        true, true,
-        window, null,
-        0, 0, 0, 0,
-        false, false, false, false,
-        0, null
-    );
-    output.output.$canvas[0].dispatchEvent(ev);
+    output.output.$canvas.mouseup();
```
kevinbarabash: Yay!

tests/shared_utils.js
```diff
@@ -1,9 +1,9 @@
 var runTestFiles = function(defaultFiles) {
-    var isPhantomJS = /PhantomJS/.test(navigator.userAgent);
+    var isHeadless = /HeadlessChrome/.test(navigator.userAgent);
 
     mocha.setup({
         ui: 'bdd',
-        reporter: isPhantomJS ? jsonReporter : 'html'
+        reporter: isHeadless ? jsonReporter : 'html'
```
kevinbarabash: I'm glad this didn't require very many changes.

testutil/gulp-mocha-runner.js
```diff
@@ -44,42 +46,73 @@ var plugin = function (options) {
...
+            async function launchChrome() {
+                return await chromeLauncher.launch({
+                    chromeFlags: [
+                        "--disable-background-timer-throttling",
+                        "--disable-default-apps",
+                        "--disable-device-discovery-notifications",
+                        "--disable-gpu",
+                        "--disable-popup-blocking",
+                        "--disable-renderer-backgrounding",
+                        "--disable-translate",
```
kevinbarabash: Not that we need these features, but I'm curious why `popup-blocking` and `translate` are disabled.

testutil/gulp-mocha-runner.js
```diff
@@ -44,42 +46,73 @@ var plugin = function (options) {
...
+                        "--headless",
+                        "--no-default-browser-check",
+                        "--no-first-run",
+                        "--autoplay-policy=no-user-gesture-required",
+                        "--allow-file-access-from-files"
+                    ],
+                    logLevel: "error"
```
kevinbarabash: This means that we don't log anything below and error?

testutil/gulp-mocha-runner.js
```diff
@@ -44,42 +46,73 @@ var plugin = function (options) {
...
+                });
+            }
+
+            const chrome = await launchChrome();
+            const protocol = await CDP({port: chrome.port});
```
kevinbarabash: I've been experimenting with puppeteer in a different context.  It's cool to see the CDP used directly.

testutil/gulp-mocha-runner.js
```diff
@@ -44,42 +46,73 @@ var plugin = function (options) {
...
-                    if (msgStr.indexOf("pass") > 0) {
-                        process.stdout.write("\x1b[32m.\x1b[0m"); // green
-                    } else if (msgStr.indexOf("fail") > 0) {
-                        process.stdout.write("\x1b[31mF\x1b[0m"); // red
-                        failures.push({
-                            fullTitle: "unknown",
-                            message: "unknown"
-                        });
-                    }
+                    messages.push(unmirror(args[0]));
```
kevinbarabash: So `mirror` is a serialization format that Chrome uses for JS objects.  It looks like sometimes we JSON strings and other times we get these mirror objects.  Could you add a comment about when we might expect each of these possible formats?

testutil/gulp-mocha-runner.js
pamelafox: Hm. I based the flags on the gulp-mocha-chrome ones, and then added the last two (autoplay and file-access). I can add a comment about that.

testutil/gulp-mocha-runner.js
pamelafox: Hm yeah, the default is info, that may be better. I'll just remove it for now, I added it when I was debugging heavily.  Reference is at https://www.npmjs.com/package/chrome-launcher for future reference.

testutil/gulp-mocha-runner.js
pamelafox: Good point, this part was confusing for me as well. Adding a comment.

Issue #545, opened 2016-01-28
<SCRIPT> tag in uppercase incorrectly generates error 


SpongeJr:
Reported on Zendesk:

> When a student opens the `<SCRIPT>` tag, they automatically are given a closing script tag `</SCRIPT>` that matches the case they entered:
> 
> `<script></script> 
> <SCRIPT></SCRIPT>`
> 
> The upper case closing tag `</SCRIPT>` immediately generates the error message
> 
> "It looks like your `<script>` tag never closes."
> 
> The parser does not appear to accept upper case closing tag `</SCRIPT>`
> 

Replicated, example can be seen on [this project](https://www.khanacademy.org/computer-programming/uppercase-script-tag-error/5220028773695488).

All caps would still be valid HTML for the`<script>` tag, so this should be accepted.

SpongeJr:
Also, I haven't found any other tags that aren't accepted in uppercase.

kevinbarabash:
@SpongeJr thanks for the report.

JettBurns14:
I found out that `<textarea>` is affected in exactly the same way, because the parser is checking the **contents** of these two specific tags, making it a unique parsing situation that could go wrong.

I've fixed this issue on my branch, will make a PR (with tests included)!

PR #660, opened 2018-07-09
Fix issues with uppercase html tag parsing

JettBurns14:
Addresses #545.

- I fixed the issue of the parser not reading the closing tag of an uppercase `<SCRIPT>` or `<TEXTAREA>` tag. Oh Noes would say the "tag never closes".
- Added tests to accommodate the changes.
- A while back the `<frameset>` and `<frame>` elements were banned, so I went ahead and added tests for them while I was at it.

Couple minor changes having to do with style were included:
- I changed the ESLint `"no-console"` rule since we are obviously using `console` a lot, and Atom's ESLint was bothering me about it. I can totally change it back if we want to keep the rule, just didn't make sense having it.
- Gulpfile now follows ESLint rules. No real reason for changing this, I just like cleaning code as I read it.

Screenshot of the demo. I made sure to hard refresh the page to make sure Oh Noes wasn't bugging out:
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,7 @@
     "max-len": [2, 80, 4, { "ignoreUrls": true, "ignorePattern": "\\brequire\\([\"']|eslint-disable" }],
     "no-alert": 2,
     "no-array-constructor": 2,
-    "no-console": 0,
+    "no-console": 2,
     "no-const-assign": 2,
     "no-debugger": 2,
     "no-dupe-class-members": 2,
--- a/.eslintrc
+++ b/.eslintrc
@@ -16,7 +16,7 @@
     "max-len": [2, 80, 4, { "ignoreUrls": true, "ignorePattern": "\\brequire\\([\"']|eslint-disable" }],
     "no-alert": 2,
     "no-array-constructor": 2,
-    "no-console": 2,
+    "no-console": 0,
     "no-const-assign": 2,
     "no-debugger": 2,
     "no-dupe-class-members": 2,
--- a/external/slowparse/slowparse.js
+++ b/external/slowparse/slowparse.js
@@ -1629,7 +1629,7 @@
     // passed the beginning `<tagname` of an HTML element.
     _parseCDATA: function(tagname) {
       var token,
-           matchString = '</'+tagname+'>',
+           matchString = `</${tagname}>`,
            text,
            textInterval = { start: 0, end: 0 },
            openTagEnd = this.domBuilder.currentNode.parseInfo.openTag.end,
@@ -1637,7 +1637,7 @@
 
       this.stream.makeToken();
       while (!this.stream.end()) {
-        if (this.stream.match(matchString, true)) {
+        if (this.stream.match(matchString, true, true)) {
           token = this.stream.makeToken();
           text = token.value.slice(0, -matchString.length);
           closeTagInterval = {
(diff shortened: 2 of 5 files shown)


.eslintrc
```diff
@@ -16,7 +16,7 @@
     "max-len": [2, 80, 4, { "ignoreUrls": true, "ignorePattern": "\\brequire\\([\"']|eslint-disable" }],
     "no-alert": 2,
     "no-array-constructor": 2,
-    "no-console": 2,
+    "no-console": 0,
```
pamelafox: Do you mean that our production code uses console? (This is a typical rule in eslint due to prevalence of engineers leaving console.log()s in their code from debugging)
How many places does the production code use console?

external/slowparse/slowparse.js
```diff
@@ -1629,15 +1629,15 @@
     // passed the beginning `<tagname` of an HTML element.
     _parseCDATA: function(tagname) {
       var token,
-           matchString = '</'+tagname+'>',
+           matchString = `</${tagname}>`,
```
pamelafox: Template literals are nice, but let's keep this PR to just the code that's needed to get this to work (versus ES upgrades).
I think it's possible we might want to merge our slowparse changes with the upstream slowparse changes, so it's best to restrict our changes to what's needed.

external/slowparse/slowparse.js
```diff
@@ -1629,15 +1629,15 @@
...
            text,
            textInterval = { start: 0, end: 0 },
            openTagEnd = this.domBuilder.currentNode.parseInfo.openTag.end,
            closeTagInterval;
 
       this.stream.makeToken();
       while (!this.stream.end()) {
-        if (this.stream.match(matchString, true)) {
+        if (this.stream.match(matchString, true, true)) {
```
pamelafox: What case does this cover? Is there a test for it?

gulpfile.js
```diff
@@ -30,11 +30,11 @@ gulp.task("templates", function() {
     gulp.src(paths.templates)
         .pipe(changed("build/tmpl", {extension: ".js"}))
         .pipe(handlebars({
-            handlebars: require("handlebars")
+            handlebars: require("handlebars"),
```
pamelafox: Hm, these are nice changes, but I have a separate branch which ports live-editor to webpack, which will get rid of a bunch of those code anyway. Can you stash this for a future PR? That'll keep this PR to one feature only, and be better timing after mine is merged.

.eslintrc
JettBurns14: Aha, that makes sense. As far as I saw the production code does not use console, I mainly saw this used in Gulpfile, which is local. My next commit can revert this.

external/slowparse/slowparse.js
JettBurns14: Understood.

gulpfile.js
JettBurns14: Sure, I can stash. Excited for the merge!

external/slowparse/slowparse.js
JettBurns14: I'm not familiar enough with Slowparse to know _exactly_ what it all does, but the comments above the `Stream.prototype.match()` method say
> caseFold can be set to `true` to make the match case-insensitive.

That's what this does here, and immediately fixed the problem when I passed `true` to it. I added tests that try different cases [here](https://github.com/Khan/live-editor/pull/660/files#diff-838964eb6b468cbc568d777be1895122R96), they passed.

pamelafox: Ah, so there was previously a change that involved calling .toLowerCase(), and I thought that was how you fixed it. Did you realize you only needed to change line 1640 to fix it, and then reverted the .toLowerCase() calls?

JettBurns14: Haha, exactly. I had initially tried using .toLowerCase() to no avail, got it working by changing line 1640, then tested some more this morning and found the lowercasing wasn't even needed.

Sorry for any confusion, this PR should have been a lot cleaner... at least I'll learn from it!

Issue #640, opened 2017-06-27
comma separated var dec/uninitialize in for loop gets dropped


EricBalingit:
possibly related to #603

    for ( var i = 0, x; i < 10; i = i + 1 ) {
        x = i * 1;
    }

Oh noes! Cannot read property 'type' of null

https://www.khanacademy.org/computer-programming/no-joy/5467046852493312

PR #661, opened 2018-07-10
Babyhint param count updates

EthanLuisMcDonough:
### High-level description of the changes
This pull request ~addresses issue 640 and~ fixes the param counts present in [babyhint.js](https://github.com/Khan/live-editor/blob/master/js/output/pjs/babyhint.js)

### Are there performance implications for this change?
As far as I'm aware, no.

### Have you added tests to cover this new/updated code?
Yes.

### Risks involved
I don't see any risks present in these changes.

### Are there any dependencies or blockers for merging this?
No.

### How can we verify that this change works?
I've added tests.

PR #663, opened 2018-07-12
Fix issue #640

EthanLuisMcDonough:
### High-level description of the change
Addresses #640.  This is the other half of #661.

No.

Yes.

### Risks involved
None that I know of.

No.

I added a test.


kevinbarabash:
I don't know why travis thinks that:
```
Promise Rejection:  { Error: connect ECONNREFUSED 127.0.0.1:34281
    at Object._errnoException (util.js:992:11)
    at _exceptionWithHostPort (util.js:1014:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1186:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 34281 }
```
means the tests have passed.

EthanLuisMcDonough:
The fork has been synced.  All tests pass on my machine.

pamelafox:
I checked the PR out locally, confirmed tests passed, and retriggered the TravisCI build, which does pass now. I took a while on this PR because I didn't write the AST transform code and wanted to understand it better. Your change looks like a change for the better since it avoids a runtime error and it is in line with other filter()s in other parts of the code. 

(I don't fully understand why we attach __env in some cases and not all, but your code doesn't seem to introduce additional complexity around that question)

PR #665, opened 2018-07-20
Uses requestAnimationFrame to prevent frozen scripts while seeking

scotchfield:

We've had some reports about the screen locking up while seeking ahead. This is caused by processing commands and building the cache, which trap us in the single thread and prevent other behaviour.

This approach uses `window.requestAnimationFrame` to process a few commands and cache entries, then see if we need to update the browser before trying another pass.

I played with the value of `animationStep` and this seemed okay! It might make sense to show more or less though, so that would be a value to play around with. There's also some place where it's possible to generate bad cache entries if you continue to seek while it's still processing stuff. You can see this if you visit a Khan page like http://localhost:8081/computing/computer-programming/programming/drawing-basics/p/making-drawings-with-code, seek ahead to the end, and while it's processing, start scrolling around the scrollbar. That might be a problem in our Khan `webapp` though; I wasn't sure if I missed returning out of somewhere here in `live-editor` while seeking like in the `play` function, or if this is coming from the scroll bar in `webapp`.

I haven't profiled this code, but it leads to more responsive seeking in webapp. As a result, it might be slightly slower because of the extra work.

No new tests, but I ran the old tests and they all passed.

### Risks involved

It's possible to mess up the cache if the seek is interrupted or the editor state is changed from outside while seeking. I don't think this solution is complete yet as a result; there's something that could still affect the result during playback.

No.

My testing so far has been with the test suite here and by verifying on live examples in the Khan Academy webapp on `localhost`.

### DIFF
--- a/js/shared/record.js
+++ b/js/shared/record.js
@@ -21,6 +21,10 @@ window.ScratchpadRecord = Backbone.Model.extend({
         // Array of command arrays for each chunk. Only holds chunks that
         // have been saved.
         this.allSavedCommands = [];
+
+        // True when we actively seeking to a new position and potentially
+        // building the cache.
+        this.seeking = false;
     },
 
     setActualInitData: function(actualData) {
@@ -141,6 +145,12 @@ window.ScratchpadRecord = Backbone.Model.extend({
     // Seek to a given position in the playback, executing all the
     // commands in the interim
     seekTo: function(time) {
+        if (this.seeking) {
+            return;
+        }
+
+        this.seeking = true;
+
         // Initialize and seek to the desired position
         this.pauseTime = (new Date()).getTime();
         this.playStart = this.pauseTime - time;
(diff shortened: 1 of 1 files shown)


js/shared/record.js
```diff
@@ -188,13 +198,27 @@ window.ScratchpadRecord = Backbone.Model.extend({
...
-        for (var i = cacheOffset; i <= seekPos; i++) {
-            this.runCommand(this.commands[i]);
-            this.cache(i);
+        let currentOffset = cacheOffset;
+        const buildCache = () => {
+            const animationStep = 100;
+            const steps = Math.min(animationStep, seekPos - currentOffset + 1);
+            const newOffset = currentOffset + steps;
+
+            // console.log(`${currentOffset} / ${newOffset} / ${seekPos}`);
```
pamelafox: Is this intended to be still in the code?

js/shared/record.js
```diff
@@ -188,13 +198,27 @@ window.ScratchpadRecord = Backbone.Model.extend({
             this.cacheRestore(-1 * this.seekCacheInterval);
         }
 
-        // Execute commands and build cache, bringing state up to current
-        for (var i = cacheOffset; i <= seekPos; i++) {
-            this.runCommand(this.commands[i]);
-            this.cache(i);
+        let currentOffset = cacheOffset;
```
pamelafox: Could you add a comment above this section about why it uses RAF? (i.e. a short version of your PR descrip)

js/shared/record.js
scotchfield: Oops, no! Sorry about that. I'll remove it.

js/shared/record.js
scotchfield: For sure! Adding it now.

MatthiasSaihttam:
If I'm seeing reading right, this PR didn't include building these changes, meaning the build files are out of date.

pamelafox:
Yes, we just realized that. I will add build as a pre-commit step in the future, to prevent them getting out of date.

PR #669, opened 2018-08-19
Refactor live-editor to use webpack and React

pamelafox:

This is a very large refactor of live-editor. The original intention was to move from bower to webpack. However, I found that I was spending an inordinate amount of time getting the last 1% of the old technology working with webpack, and that it would be easier to modernize the whole UI stack.
I apologize profusely for the massiveness of this PR.

The big changes:
* **Bower -> Webpack**: It now uses two webpack files, one for local demos and testing, the other for integration with KA webapp. The big difference is in the externals.
* **Backbone View -> React**: All of the views are now React. This was tricky as we had a lot of child<->parent communication that had to work in Reactland, so you'll see a lot of children calling callbacks. I also used the event-sending technique where the parent timestamps a prop and the child compares timestamps to see if its a new event.
* **CSS -> Aphrodite**: Goes along with the React change. There's a handful of CSS left, in a single file.
* **WonderBlocks UI**: This uses buttons, modals, icons, and spinners from WonderBlocks.
* **Upgraded ACE editor**: The only big change was a merging of their insert/insertLine events.
* **Upgraded SoundManager2**: No changes noticed.
* **--Handlebars/Bootstrap/jQuery/jQueryUI**: No longer needed, thanks to React and WB. 
* **--visibly**: no longer needed as of IE11, last browser we fully support.
* **--poster/testee**: not used by the deployed code, AFAICT.

For this refactor, I intentionally left out parts that are currently not used by Khan Academy webapp:
* **Structured-Blocks**: that was removed from webapp a while ago, due to bugs and no time for maintenance. The trend continues! I left the files around and moved the CSS into its own file, so it is possible to resurrect it.
* **Debugger**: That has always been behind a flag, so I didn't spend the time to get it working. It probably wouldn't be hard to get working again, as compared to StructuredBlocks.
* **Document editor**: I did get this working, but I'm debating whether to go through the effort of bringing CKeditor into this repo, as that's what webapp uses (and it really should be in this repo). It's not used by any curriculum on KA though, so I'm going to grep some analytics to see who's using it.
* **Recording functionality**: I left the code in, but did not go through effort to get it to work, as it likely barely worked before. The talk-through playback functionality should still work, just not making new recordings.

I also made some infrastructure changes, mostly by copying WonderBlocks:
* **ESlint** (with KA's eslintrc, plus some nice import ordering rules)
* **Prettier** (with old files in prettierignore, to minimize changes in this PR)
* **Pre-commit** calls lint, pretty-quick, build, test

To make this much-too-large PR easier to grok, I've made a doc describing the high level architecture:
https://docs.google.com/document/d/1aa9r4VJAJZ8KDbnJ634BTsuEm267gBveViaIlZcBAEc/edit#

It may be easier to review this if you read that first and then CTRL+F for specific components to review.
If you notice any React style issues that occur multiple times, you can mention them once, and I can try to fix all the occurrences. 

I still have a few UI issues to fix and will likely add a few more prop callbacks as I continue my webapp integration, but I believe those will be small changes to the code.

Likely, due to React versus Backbone. There are also less dependencies now, so the code bundles should hopefully be smaller on webapp.
The KA webapp code still uses jQuery/jQueryUI though, so we'd need to upgrade that before we'd get the full benefit of having less dependencies in live-editor.

I ported the previous tests - the non-UI tests didn't change much, but the UI tests had to change a bunch, as they were highly implementation dependent. 
I realized a number of the tests were actually single-component tests, so I moved them into the js/ folder to sit alongside the component they test (like pjs-code-injector.test.js).

The only tests that didn't make it were 2 tests that checked the anti-undo mechanism, as they involved a lot of gross reaching into the code. I manually verified the anti-undo and it seems to work as expected.

You can run tests in the browser at localhost/tests or in Chrome headless via `npm run test`.

I did not add any fixtures, like we have in KA webapp. I'm not sure if that's a useful thing to have in live-editor, given we don't have a React sandbox here?

### Risks involved

This is a huge change, it's *very* likely for this to introduce bugs. I plan to deploy a ZND with the change in webapp and do a number of manual tests on that. There's a QA person that's been running manual tests on scratchpads this summer, and they may be able to test this change.
I imagine there will still be some bugs that pop up after deployed, so I will be keeping ane eye out for those.

* StructuredJS PR: approved, I just need to check the formatting and merge.
* Webapp PR: I have a bit more work to get it fully working in webapp, and then I'll send a PR for that. I wanted to send this earlier since this is a larger change.

With hundreds of manual tests! :) 

### Screenshots

Here are various screenshots of the new UI.
screen shot 2018-08-19 at 10 09 37 am
screen shot 2018-08-19 at 10 08 49 am
screen shot 2018-08-19 at 10 07 59 am
screen shot 2018-08-19 at 10 07 50 am
screen shot 2018-08-19 at 10 07 36 am
screen shot 2018-08-19 at 10 07 12 am
screen shot 2018-08-19 at 10 07 05 am


Planetarianproductions:
Amazing!

pamelafox:
Update: this PR is too large. To make it easier to review, I will be breaking this into smaller PRs where possible (like the visibly PR, just sent).

pamelafox:
I'm closing this PR to make it clear I'm not asking for a review on it, and am instead sending smaller PRs for what's covered in this PR.

Issue #260, opened 2015-02-24
Infinite loop detection in webpages pops up after tab switch


pamelafox:
This code can result in an infinite loop error:
```
  <!DOCTYPE html>
  <html>
  <head>
      <meta charset="utf-8">
      <title>Animation starter</title>
  </head>
  <body>
  <div>
  <img src="https://www.kasandbox.org/programming-images/creatures/OhNoes.png">
  <h1>Oh noes, the world will end in
      <span id="countdown">500</span>
      seconds!
  </h1>
  
  </div>
  
<script>
// Step 1. What element do we want to animate?
var countdown = document.getElementById("countdown");
// Step 2. What function will change it each time?
var countItDown = function() {
  countdown.innerHTML = countdown.innerHTML - 1;  
};
// Step 3: Call that on an interval
window.setInterval(countItDown, 1000);
</script>

  </body>
  </html>
```

Investigation from john:
It only happens if you tab away from the page with the code and leave it in the background. After some amount of time (haven't pinned it down yet) this code doesn't fire at the right time:

setTimeout(function() {
    this.branchStartTime = 0;
}.bind(this), 0);
since the page is in the background the timers aren't run in the right order. The hiccup looks something like this:
https://www.dropbox.com/s/ulq0um8r7m4f3ws/Screenshot%202015-01-28%2013.58.40.png?dl=0

One thought is to use the page visibility API to not do checking while the page is in the background:
https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API

kevinb7:
I had a look at the loop protector code.  Using page visibility to disable/enable the loop protector seems like it would work and wouldn't be too much effort.  I wonder if something similar to this would help with infinite loops in pjs environment.

kevinb7:
Page Visibility API isn't supported in IE9, but I found an polyfill that I'm going to try out: https://gist.github.com/addyosmani/1122546.

kevinb7:
Fixed by https://github.com/Khan/live-editor/commit/6b3b5e505179649c1a291d1dcfe76872b463ee0a.

PR #670, opened 2018-08-22
Remove visibly, replace with Page Visibility API

pamelafox:

This replaces the visibly submodule with the Page Visibility API, now supported in all relevant browsers. See: https://caniuse.com/#feat=pagevisibility

One less module, so smaller size! (Also, visibly was likely polling. Native is better).

No, I ran previous tests and ran the manual test.

### Risks involved

There could be a subtle difference between visibly and the visibility API? Hopefully not, since visibly was a polyfill.

No.

- copy/paste the example from #260 into the demo app
- hide the browser or switch to a different tab and wait for ~2 minutes
- return to the demo page and verify that the script has continued to run uninterrupted

Taken from the original visibly PR: https://github.com/Khan/live-editor/commit/6b3b5e505179649c1a291d1dcfe76872b463ee0a

### DIFF
--- a/build/js/live-editor.shared.js
+++ b/build/js/live-editor.shared.js
@@ -136,6 +136,10 @@ window.ScratchpadRecord = Backbone.Model.extend({
         // Array of command arrays for each chunk. Only holds chunks that
         // have been saved.
         this.allSavedCommands = [];
+
+        // True when we actively seeking to a new position and potentially
+        // building the cache.
+        this.seeking = false;
     },
 
     setActualInitData: function setActualInitData(actualData) {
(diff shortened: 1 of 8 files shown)


build/js/live-editor.shared.js
```diff
@@ -136,6 +136,10 @@ window.ScratchpadRecord = Backbone.Model.extend({
         // Array of command arrays for each chunk. Only holds chunks that
         // have been saved.
         this.allSavedCommands = [];
```
pamelafox: Build files were left out of the last PR, which is why these changes are here. I've sent a separate PR to Scott for them though, will merge that first.

PR #671, opened 2018-08-22
Update build files for RAF change

pamelafox:

This updates the build files, which were left out of the requestAnimationFrame change: https://github.com/Khan/live-editor/pull/665

There still needs to be a webapp PR to update the live-editor submodule.

See https://github.com/Khan/live-editor/pull/665


pamelafox:
Confirmed: tests are running against compiled code. (Both in master and in webpack branch)

PR #672, opened 2018-08-22
Port from jQuery to DOM in WebpageTester, fix the associated tests

pamelafox:

This replaces jQuery with native DOM in WebpageTester (and an associated line in WebpageOutput).
I discovered that the tests weren't actually testing the results of the validation tests (as they were returning true even when my port was clearly faulty), so I also fixed the tests. 

Teensy bit better due to not passing through jQuery.

I did not add tests, just fixed the existing ones and verified they went through the code path.

### Risks involved

Usual risk of jQuery to native, any differences.

This PR has a build file from the PR that accidentally had no build files, so I should merge that first.

On KA, do a challenge in HTML/JS course, like https://www.khanacademy.org/computing/computer-programming/html-css-js/html-js-dom-access/p/challenge-query-modernizer
Verify you can pass and fail steps.

### DIFF
--- a/js/output/webpage/webpage-tester.js
+++ b/js/output/webpage/webpage-tester.js
@@ -24,10 +24,10 @@
             var parser = new DOMParser();
             var doc = parser.parseFromString(userCode, "text/html");
 
-            $(doc).find("script").each(function (index, scriptElement) {
+            doc.querySelectorAll("script").forEach((scriptElement) => {
                 this.testContext.allScripts += scriptElement.innerHTML;
                 this.testContext.allScripts += "\n";
-            }.bind(this));
+            });
 
             this.curTask = null;
             this.curTest = null;
@@ -162,8 +162,8 @@
                 var expected = structure[selector];
                 // TODO(jeresig): Maybe find a way to do this such that we can run
                 // it in a worker thread.
-                var numFound = $(selector, this.testContext.$docSP).length;
-                if (expected === 0 && numFound !== 0 || numFound < expected) {
+                var numFound = this.testContext.docSP.querySelectorAll(selector).length;
+                if ((expected === 0 && numFound !== 0) || numFound < expected) {
                     return {success: false};
                 }
             }
--- a/tests/output/webpage/assert_test.js
+++ b/tests/output/webpage/assert_test.js
@@ -64,7 +64,8 @@ describe("Challenge Assertions - HTML", function() {
     assertTest({
         title: "Doing the step with no errors",
         code: "<p id='foo'><div></div></p>",
-        validate: divTest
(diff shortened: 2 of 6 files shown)


js/output/webpage/webpage-tester.js
```diff
@@ -162,8 +162,8 @@
                 var expected = structure[selector];
                 // TODO(jeresig): Maybe find a way to do this such that we can run
                 // it in a worker thread.
-                var numFound = $(selector, this.testContext.$docSP).length;
-                if (expected === 0 && numFound !== 0 || numFound < expected) {
+                var numFound = this.testContext.docSP.querySelectorAll(selector).length;
+                if ((expected === 0 && numFound !== 0) || numFound < expected) {
```
kevinbarabash: nice

tests/output/webpage/assert_test.js
```diff
@@ -170,20 +175,23 @@ describe("scriptTest tests", function() {
         title: "scriptTest reports success for matching code",
         code: "<div><script>var x = 4;</script></div>",
         validate: xTest,
-    }); 
+    });
 
     assertTest({
         title: "scriptTest reports failure for not matching code",
         code: "<div></div>",
```
kevinbarabash: I wish the titles were more descriptive.  I assume that we don't like empty tags or empty tags that aren't self-closing.

tests/output/webpage/test_utils.js
```diff
@@ -116,23 +116,36 @@ var warningTest = function(title, code, warnings) {
...
+                expect(testResults[0].state).to.not.equal("fail");
+            } else {
                 expect(testResults).to.not.equal([]);
                 expect(testResults[0].state).to.be.equal("fail");
-                expect(testResults[0].results[0].meta.alsoMessage)
-                    .to.be.equal(options.reason);
+                // If specific reason given, verify it matches:
+                if (options.reason !== "fail") {
+                    expect(testResults[0].results[0].meta.alsoMessage)
+                        .to.be.equal(options.reason);
```
kevinbarabash: I guess if we get an error here the `expect` will still fail since we won't get the right `testResults`.

tests/output/webpage/assert_test.js
pamelafox: Ah, this is a test result failure, not a lint result failure. Yeah, I did totally reword a bunch of test titles in my big PR, as I prefer verbosity in test descriptions. (There are a bunch of one-word test titles, uck). There might be a PR for that upcoming. :)

pamelafox: And by test, I mean validation test, challenge test. The word "test" gets confusing here.

PR #673, opened 2018-08-22
Adding precommit hook for test and build

pamelafox:

This adds a pre-commit hook that builds, then tests (as tests use built files), and then calls "git add build" to add any changed files. It was handy that there was existing un-committed build files to test this on in the repo, as you can see they show up in this PR!

This should prevent future situations where we forget to run build & test before committing.

Commits will be slower :)

### Risks involved

There's no review stage for reviewing the built files. That might be nice theoretically, but I'm not sure how to do that / how needed it is, since you can review in the PR as well. Let me know if you think that's needed though.

FYI @scotchfield

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "dist/live-editor.js",
   "scripts": {
     "build": "gulp build",
-    "test": "gulp test"
+    "test": "gulp test",
+    "build:add": "git add build/"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


kevinbarabash:
We're running the tests against the built files though, right?

package.json
```diff
@@ -44,12 +45,18 @@
...
     "run-sequence": "^0.3.6",
     "sinon": "^1.10.3",
     "through2": "^2.0.0",
     "vinyl-sourcemaps-apply": "^0.1.4"
-  }
+  },
+  "pre-commit": [
+    "build",
+    "test",
+    "build:add"
```
kevinbarabash: Sometimes this takes a while do run locally so we could add a check in travis.yml to verify that no files have been added/changed after running these files on CI.

pamelafox: Oh, true, we have checks from Travis as well! I've been using the pre-commit test hook for the last week locally on my branch and haven't minded the test wait. But if it gets too slow with more tests (or feels prohibitively slow for others), yep, you're right, we can defer to Travis with that file check.

PR #675, opened 2018-08-22
Fix misuse of let/var in LoopProtect

pamelafox:

This fixes mis-use of let and const in loop-protect.js, changing lets to consts and vice versa.
As a general rule, I'm not currently fixing no-var and prefer-const in old files. However, eslint did point out that this file reversed let/const in a few places, and that seemed worth fixing.

No, just ran old ones.

### Risks involved

None foreseen.

No.

Write a program with a while loop, live on the wild side! :)

### DIFF
--- a/js/output/shared/loop-protect.js
+++ b/js/output/shared/loop-protect.js
@@ -62,7 +62,8 @@ window.LoopProtector.prototype = {
             }
             this.loopCounts[location] += 1;
         }
-        let now = new Date().getTime();
+        let error;
+        const now = new Date().getTime();
         if (!this.branchStartTime) {
             this.branchStartTime = now;
             setTimeout(function () {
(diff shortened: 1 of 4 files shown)


js/output/shared/loop-protect.js
```diff
@@ -71,7 +72,7 @@ window.LoopProtector.prototype = {
         } else if (now - this.branchStartTime > this.timeout) {
             if (this.visible) {
                 if (!this.reportLocation) {
-                    let error = new Error("KA_INFINITE_LOOP");
+                    error = new Error("KA_INFINITE_LOOP");
```
pamelafox: I can also change both to const instead of introducing a let above, either way works.

kevinbarabash: Let's change both to `const`.

pamelafox:
Changed to consts, merging.

PR #676, opened 2018-08-22
Use native DOM in PJSResourceCache

pamelafox:

This replaces jQuery with native DOM in PJSResourceCache.
(It leaves $.Deferred, as that requires a promise-polyfill, so that will be a separate PR).

I verified by inspecting the created <div> and the network requests.

Teensy bit better.

No.

### Risks involved

None forseen.

No.

Put code like this in a program:
image(getImage("avatars/leaf-orange"), 10, 20);
playSound(getSound("rpg/metal-clink"));
You should see the resources loaded immediately when you load the program page, and no 404 errors in console.

### DIFF
--- a/js/output/pjs/pjs-resource-cache.js
+++ b/js/output/pjs/pjs-resource-cache.js
@@ -7,6 +7,13 @@ function PJSResourceCache(options) {
     // Insert the images into a hidden div to cause them to load
     // but not be visible to the user
     if (!this.imageHolder) {
+        this.imageHolder = document.createElement("div");
+        this.imageHolder.style.height = 0;
+        this.imageHolder.style.width = 0;
+        this.imageHolder.style.overflow = "hidden";
+        this.imageHolder.style.position = "absolute";
+        document.body.appendChild(this.imageHolder);
+
         this.imageHolder = $("<div>")
             .css({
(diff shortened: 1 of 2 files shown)


js/output/pjs/pjs-resource-cache.js
```diff
@@ -7,14 +7,12 @@ function PJSResourceCache(options) {
...
-                width: 0,
-                overflow: "hidden",
-                position: "absolute"
-            })
-            .appendTo("body");
+        this.imageHolder = document.createElement("div");
+        this.imageHolder.style.height = 0;
+        this.imageHolder.style.width = 0;
+        this.imageHolder.style.overflow = "hidden";
+        this.imageHolder.style.position = "absolute";
```
kevinbarabash: I wonder if `Object.assign(this.imageHolder.style, {height: 0, width: 0, overflow: "hidden", position: "absolute"})` would work.  Since this works and is already written let's go with it.

pamelafox: That does seem to work, just tried here:
http://jsbin.com/lapisetaki/edit?html,output
Don't know if it'd work in all browsers but interesting to see.
I'll leave it like this, as you say.

PR #681, opened 2018-09-13
Fix #542

### LINKED ISSUE #542: Yoko Oh Noes has a beef with literal arrays as arguments to println
`println([3, 2, 1]);`
produces the complaint "t looks like you're trying to use println(). In case you forgot, you can use it like: println(data) | ProcessingJS".

EthanLuisMcDonough:
This change makes it so that Babyhint accurately counts the number of arguments passed into a function.  Fixes #542.

`countCommas` might not be as fast as `match(/,/g)`.

Yes.  The tests I added are present in `jshint_test.js`.

### Risks involved
I can't think of any risks involved with this PR.

No.

I added tests.

### DIFF
--- a/js/output/pjs/babyhint.js
+++ b/js/output/pjs/babyhint.js
@@ -559,16 +559,17 @@ var BabyHint = {
         };
 
         var stack = [];
+        var indexStack = [];
         var commaCount = 0;
-        var lastPushIndex = 0;
 
         for (var i = 0; i < str.length; i++) {
             if (pairs.hasOwnProperty(str[i])) {
                 stack.push(str[i]);
-                lastPushIndex = i;
+                indexStack.push(i);
             } else if (revPairs.hasOwnProperty(str[i])) {
                 if (revPairs[str[i]] === stack[stack.length - 1]) {
                     stack.pop();
+                    indexStack.pop();
                 } else {
                     return {
                         error: {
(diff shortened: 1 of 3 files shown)


pamelafox:
Kevin mentioned using the AST to check this in the issue discussion. Did you attempt an AST-based approach?

EthanLuisMcDonough:
I considered it but then opted for my approach because it was more convenient.  I don't mind working on a version that uses Esprima if you think that BabyHint would benefit from such a refactor, though.

js/output/pjs/babyhint.js
```diff
@@ -641,10 +687,23 @@ var BabyHint = {
...
+
+            // push in an error if an error occurs while parsing the param list
+            if (rawCount.error) {
+                errors.push({
+                    row: lineNumber,
+                    column: index + 2 + rawCount.error.index,
+                    text: rawCount.error.message,
+                    breaksCode: false,
+                    source: "paramschecker",
+                    context: {}
```
pamelafox: The tests don't seem to cover the two possible errors from the paramschecker, from what I can tell, is that right? Can you add tests that cover those errors?

EthanLuisMcDonough: Ok, I added some more tests.

pamelafox:
This is now deployed on khanacademy.org

PR #682, opened 2018-09-14
Normalizing console statements, removing window

pamelafox:

The main point of this PR is to remove "window" from "window.console" statements, as this moves us towards easier webpack integration. We have some places where we use "console &&" before console, so I normalized to that everywhere. However, I'm happy to remove "console &&" if that's not needed.

No

No, just ran current tests.

### Risks involved

There's some risk of console && not working in all browsers, but at this point, with modern browsers, that seems less likely.
I checked webapp and we have a few different approaches there.

No

Use the PJS editor, all should work with no errors, particularly debug()

### DIFF
--- a/js/output/pjs/pjs-output.js
+++ b/js/output/pjs/pjs-output.js
@@ -437,7 +437,7 @@ window.PJSOutput = Backbone.View.extend({
                     // figure out whether we hit this case.
                     var message = i18n._("Error: %(message)s",
                         {message: errors[errors.length - 1].message});
-                    console && console.warn(message);
+                    console.warn(message);
                     this.tester.testContext.assert(false, message,
                         i18n._("A critical problem occurred in your program " +
                             "making it unable to run."));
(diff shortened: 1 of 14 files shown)


js/output/pjs/pjs-output.js
```diff
@@ -437,8 +437,7 @@ window.PJSOutput = Backbone.View.extend({
                     // figure out whether we hit this case.
                     var message = i18n._("Error: %(message)s",
                         {message: errors[errors.length - 1].message});
-                    // TODO(jeresig): Find a better way to show this
-                    this.output.$el.find(".test-errors").text(message).show();
```
pamelafox: test-errors doesn't exist anymore (hasn't in a while). We do display this in a hopper message, I believe, but I figured might as well console.warn() it as well

pamelafox:
Note: VSCode fixed up whitespace issues, so it's easier to view the diff with whitespace-ignore on.

pamelafox:
@kevinbarabash Great, I was hoping that was the case. Much simpler code now. (*Especially if you view diff with ignore-whitespace on)

kevinbarabash:
> (*Especially if you view diff with ignore-whitespace on)

I think I used it that feature once a long time ago and then forgot it existed.  It helps a lot.  I wonder if there's a way to turn it on by default.

PR #683, opened 2018-09-14
Remove PJSDebugger from code paths

pamelafox:

This removes PJSDebugger from the live code paths. It's not currently enabled, and it isn't something I had the time to port over in my webpack refactor, and it seems better to have a single PR that removes it than to remove it piecemeal. You can reference this PR in the future to bring it back into whatever the codebase looks like then.

Note that I left the actual 2 debugger files in the code tree, just like we left structured-blocks in the tree. I just tried to remove it from all actually used files.

Not really, since the debugger was never enabled, and I don't think we were bringing in its dependencies to webapp yet.

No, since I deleted code.

### Risks involved

I may have missed somewhere. I grepped for debugger, which seemed to catch a lot.

No

Everything should work as before! :)


pamelafox:
@kevinbarabash I've removed 4 debugger-related files from the build/ directory, good call:

        deleted:    build/css/live-editor.debugger.css
        deleted:    build/js/live-editor.debugger.js
        deleted:    build/js/live-editor.output_debugger_deps.js
        deleted:    build/tmpl/debugger.js

PR #686, opened 2018-09-26
Port output and tests to webpack

pamelafox:

This is a subset of PR#669 (https://github.com/Khan/live-editor/pull/669), the port to webpack/React, This PR will go into pamela-bigbranch, not into master. Subsequent PRs go also be merged into that branch, and we will not merge the full branch until everything is fully functional and tested.

This particular PR ports the output side of LiveEditor from bower to webpack. At a high level, that involves:
* Using import and export (instead of globals on window)
* Moving the copy of build assets from bower to webpack (sounds, images)
* Rewriting the demos to refer to the new bundled build files
* Rewriting the tests to use import/exports

Note that the Output JS files are all written in a large number of ways, so in some files, I exported at bottom, in others, I exported at the top - I typically picked whatever was the minimal change. I am happy to use a more consistent style, it'll probably just mean a larger diff.

The bundle sizes are different, as there are less bundles. So that could be good or bad, depending how you measure it. 

No, I ported the existing output tests over. All of the output tests pass!

### Risks involved

I wrote this PR in a way that works with the webapp kake setup. It's possible that it won't be as compatible with the webpack webapp port - hard for me to say. This does get us much closer.

This is not a fully functional branch. The output works, but the input needs to be ported, as there are files shared between the two.

The output tests all pass, and the demos work.


js/workers/pjs/jshint-worker.js
```diff
@@ -6,15 +10,6 @@ self.onmessage = function(event) {
...
         init = true;
-
-        importScripts(event.data.externalsDir +
-            "es5-shim/es5-shim.js?cachebust=" + (new Date()).toDateString());
-
-        importScripts(event.data.jshintFile +
-            "?cachebust=" + (new Date()).toDateString());
-
-        importScripts(event.data.externalsDir +
-            "underscore/underscore.js?cachebust=" + (new Date()).toDateString());
```
kevinbarabash: Why can we get rid of these `importScripts`?

js/workers/pjs/test-worker.js
```diff
@@ -1,39 +1,21 @@
...
-        importScripts(event.data.externalsDir +
-            "structuredjs/external/esprima.js?cachebust=" + date);
-        importScripts(event.data.externalsDir +
-            "underscore/underscore.js?cachebust=" + date);
-        importScripts(event.data.externalsDir +
-            "structuredjs/structured.js?cachebust=A" + date);
-
-        // Bring in the testing logic
-        importScripts("../shared/output-tester.js?cachebust=" + date);
-        importScripts("./pjs-tester.js?cachebust=" + date);
```
kevinbarabash: Why are we able to get rid of all of these `importScripts`?

package.json
```diff
@@ -21,11 +21,17 @@
   "devDependencies": {
     "async": "^0.9.0",
     "babel-core": "5.6.18",
```
kevinbarabash: This is a pretty old version of `babel-core`.  We should probably use the most recent version of the `6.x` line.

package.json
```diff
@@ -52,7 +58,25 @@
...
+  },
+  "dependencies": {
+    "ace-builds": "1.3.3",
+    "backbone": "^1.0.0",
+    "bootstrap": "^3.1.1",
+    "es5-shim": "^3.4.0",
+    "escodegen": "^1.10.0",
+    "escope": "^1.0.0",
+    "esprima": "^4.0.0",
+    "handlebars-loader": "^1.7.0",
```
kevinbarabash: Do we still need the "handlebars" dependency if we're using "handlebars-loader"?

tests/output/pjs/ast_transform_test.js
```diff
@@ -1,3 +1,17 @@
+import PJSCodeInjector from "../../../js/output/pjs/pjs-code-injector.js";
+
+const getCodeFromOptions = function(code) {
```
kevinbarabash: Where is this used?  Did you maybe forget an `export`?

tests/output/pjs/test_utils.js
```diff
@@ -43,20 +43,16 @@ var runTest = function(options) {
...
             soundsDir: "../../../sounds/",
-            jshintFile: "../../../build/external/jshint/jshint.js",
             loopProtectTimeouts: {
                 initialTimeout: 1000,
                 frameTimeout: 500
             },
         });
 
-        // Switch to the Scratchpad's version
-        output.config.switchVersion(options.version);
```
kevinbarabash: Why can we get rid of this?

tests/output/sql/index.html
```diff
@@ -9,22 +9,25 @@
...
+        <script src="../../../build/js/live-editor.i18n.js"></script>
+        <!-- Exposes html2canvas as global -->
         <script src="../../../build/js/live-editor.output_sql_deps.js"></script>
-        <script src="../../../build/js/live-editor.output_sql.js"></script>
+        <!-- Exposes SQL as global -->
+        <script src="../../../external/sqlitejs/sql.js"></script>
+        <!-- Mocha, sinon, and test JS -->
         <script src="../../../node_modules/mocha/mocha.js"></script>
         <script src="../../../node_modules/expect.js/index.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
```
kevinbarabash: Were we not using sinon before?

tests/output/sql/test_utils.js
```diff
@@ -86,14 +94,6 @@ var assertTest = function(options) {
...
             } else {
                 expect(errors).to.not.equal([]);
-                if (options.jshint) {
-                    expect(errors[0].lint).to.be.ok();
-                    expect(errors[0].lint.reason)
-                        .to.be.equal(options.reason);
-                } else {
-                    var $html = $("<div>" + errors[0].text + "</div>");
-                    expect($html.text()).to.be.equal(options.reason);
-                }
```
kevinbarabash: I'm glad this can go away now, but I don't understand why it was needed in this file in the first place since it's used by the sql tests.

tests/output/pjs/iframe.html
```diff
@@ -10,24 +10,23 @@
...
-        <script src="test_utils.js"></script>
         <script src="../../shared_utils.js"></script>
         <script>
             runTestFiles([
-                "assert_test.js",
-                "ast_transform_test.js",
-                "async_test.js",
-                "jshint_test.js",
-                "output_test.js",
+                "../../../build/js/live-editor.tests_output_pjs_jshint.js",
```
kevinbarabash: I'm guessing this change is because we can't run the source code inside phantomjs anymore, right?

tests/output/webpage/test_utils.js
```diff
@@ -1,29 +1,34 @@
...
-        options.version = ScratchpadConfig.prototype.latestVersion();
-    }
+import LiveEditorOutput from "../../../js/output/shared/output.js";
+import WebpageOutput from "../../../js/output/webpage/webpage-output.js";
+
+LiveEditorOutput.registerOutput("webpage", WebpageOutput);
 
-    var displayTitle = options.title +
-        " (Version: " + options.version + ")";
+export function createLiveEditorOutput(extraOptions) {
```
kevinbarabash: FYI, you can use default args now, e.g.
```
function createLiveEditorOutput(extraOptions = {})
```

webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+                        path.resolve(__dirname, "tmpl/helpers")
+                    ]
+                }
+            },
+            {
+                test: require.resolve("underscore"),
+                use: [
+                    {
+                        loader: "expose-loader",
+                        options: "_",
```
kevinbarabash: Can you add a comment about why we're doing this for `underscore`?

webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+    },
+    optimization: {
+        minimize: false,
+    },
+    node: {process: true},
+    mode: "development",
+    module: {
+        rules: [
+            {
+                test: /\.(js|jsx)$/,
```
kevinbarabash: `/\.js$/` should be okay since I think we want to use `.js` for files containing JSX as well.

webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+                exclude: /(node_modules|external|tests|\.(test)\.(js|jsx)$)/,
```
kevinbarabash: Does `/tests/*.test.js` need to be excluded?  It looks like our test files actually include `_test.js` and need to be transformed anyways.

webpack.config.js
```diff
@@ -0,0 +1,104 @@
+const path = require("path");
+
+const webpack = require("webpack");
+const CopyWebpackPlugin = require("copy-webpack-plugin");
+
+const config = {
+    performance: {hints: false},
```
kevinbarabash: I'm not familiar with this option, what does it do?

webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+    entry: {
+        i18n: ["./js/shared/i18n.js"],
+        output_pjs_deps: [
+            "./external/processing-js/processing.js",
+            "./external/jshint/jshint.js",
+        ],
```
kevinbarabash: This is not very webpack-y but it does what we need.

webpack.config.js
```diff
@@ -0,0 +1,104 @@
...
+            fs: true,
+            // Bundling this results in an asm error
+            "sql.js": "SQL",
+            // Separated into its own build file (to mirror webapp)
+            i18n: "i18n",
+        },
+    ],
+    output: {
+        path: path.resolve(__dirname, "build/js"),
+        filename: "live-editor.[name].js",
```
kevinbarabash: nice

kevinbarabash:
I'm pretty sure that we have automated tests for the infinite loop checker, but if my memory's failed me and we don't it'd be good to manually check that that still works.

demos/simple/index.html
```diff
@@ -64,7 +64,7 @@ <h1>Live Editor Example</h1>
         height: 400,
         editorHeight: "80%",
         autoFocus: true,
-        workersDir: "../../build/workers/",
+        workersDir: "../../build/",
```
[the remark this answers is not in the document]
pamelafox: The workers are now generated as webpack entry point, so they end up in builld/js like all the others. I could get them back in workers by changing the output option to modify the folder of the built bundles, but I haven't seen a need for workers being in their own folder yet.

js/output/pjs/pjs-code-injector.js
```diff
@@ -1076,13 +1091,16 @@ class PJSCodeInjector {
...
+                var findWhere = function(array, criteria) {
+                    return array.find(item => Object.keys(criteria).every(
+                        key => item[key] === criteria[key]));
+                };
                 var loadSound = function(filename) {
                     return new Promise((resolve) => {
                         var audio = document.createElement("audio");
                         var parts = filename.split("/");
 
-                        var group = _.findWhere(OutputSounds[0].groups, { groupName: parts[0] });
```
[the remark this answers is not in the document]
pamelafox: Yes, I think so, given we only use parts of it in webapp, and that many things are now available in ES6. I did originally make a bunch of lodash/underscore removal related changes in my big branch, but then realized that wasn't the P1 change we needed and reverted them. I kept the change here since this exported code should have minimal dependencies.

js/output/pjs/pjs-output.js
```diff
@@ -1,4 +1,14 @@
-window.PJSOutput = Backbone.View.extend({
+const $ = require("jquery");
```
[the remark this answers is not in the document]
pamelafox: Changing to imports - that's because I was using require() at the pre-React stage of the PR and I copy-pasta'd.

js/output/pjs/pjs-output.js
```diff
@@ -17,7 +27,7 @@ window.PJSOutput = Backbone.View.extend({
         this.output = options.output;
 
         this.tester = new PJSTester(_.extend(options, {
-            workerFile: "pjs/test-worker.js"
+            workerFile: "js/live-editor.test_worker.js",
```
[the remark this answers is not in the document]
pamelafox: The worker files are now built files (entry points) which contain all of their requirements. The new filename is the built file.

js/output/pjs/pjs-utils.js
```diff
@@ -1,4 +1,4 @@
-window.PJSUtils = {
+const PJSUtils = {
```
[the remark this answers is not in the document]
pamelafox: I've identified three files in this diff that would be better done that way:
- Pjs-utils
- Pjs-ast-transforms
- Ast-builder

I'm happy to make that change in this diff, if you're OK with the larger diff. (Sometimes big whitespace changes throw off the diffing so the PR looks bigger, I don't know how it'll render that change). Those files are minimally changed otherwise.

js/output/shared/output-tester.js
```diff
@@ -136,7 +142,7 @@ OutputTester.prototype = {
         if (!code) {
             return true;
         }
-
+        code = code.replace(/\$\._/g, "i18n._");
```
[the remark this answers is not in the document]
pamelafox: Good call, adding the comment.

js/workers/pjs/jshint-worker.js
pamelafox: We now load in the built worker file, so that entry point includes es5-shim and JSHint.
JSHint never used underscore. It looks like your debugger PR added underscore, and then we didn't later remove it:
https://github.com/Khan/live-editor/commit/10f0c17a3c410b9b995b9253097bf8ab8e6821a6

js/output/pjs/pjs-code-injector.js
```diff
@@ -1045,6 +1058,8 @@ class PJSCodeInjector {
 
         // TODO(kevinb) generate this code once (once webpack is in place)
         helperCode += `var resources = ${JSON.stringify(resources)};\n`;
+        // We rename it to OutputSounds2 so that webpack doesn't rewrite it
+        helperCode += `var OutputSounds2 = ${JSON.stringify(OutputSounds)};\n`;
```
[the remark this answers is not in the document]
pamelafox: Good question. It doesn't overwrite this name, it rewrites the later reference, on line 1103 (the object we call findWhere on). I'll update the comment.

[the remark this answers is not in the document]
pamelafox: You likely have a better idea for how to do this, given your TODO about improving once we use webapp. Given this is currently not in use by webapp, my tendency is to defer that.
(But let me know if you have a better quick fix)

js/output/pjs/pjs-output.js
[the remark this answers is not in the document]
pamelafox: Actually, I'm changing it to use a separate webapp config for the workers (as the externals are different), so I'll change it back to /workers.

package.json
pamelafox: Changing to "babel-core": "^6", is that right? I notice wonder-blocks using 7

package.json
pamelafox: Good Q. We import handlebars/runtime in SQL in order to register a helper dynamically, so we still use it. FOR NOW... BUM BUM BUM.

tests/output/pjs/ast_transform_test.js
pamelafox: Great catch. It's used a bunch later in the file. I'm changing it to import from pjs-utils instead. I'm confused as to why I pasted a copy in the file instead.

tests/output/pjs/iframe.html
pamelafox: It was easier for each of the tests to be a built file, including all its dependencies. Otherwise I'd need to do a script include that exposed the relevant classes, right?
(Perhaps there's another way that I'm missing)

tests/output/pjs/test_utils.js
pamelafox: Good question. I could not find any PJS tests that specified the version option, and it's not documented, and ScratchpadConfig() isn't my favorite class to have a dependency on (it's a Backbone model), so I happily did away with it here.

tests/output/sql/index.html
pamelafox: Good catch, we still don't use Sinon for SQL tests. This was copy/pasta, removing it!

tests/output/sql/test_utils.js
pamelafox: Generally the test_utils files have a lot in common with each other across pjs/webpage/sql, so I imagine the original was copy/pasta and we didn't notice that there was unused code. (If we had a coverage tool, we might identify such areas)

tests/output/webpage/index.html
```diff
@@ -9,24 +9,24 @@
...
+
+        <!-- Exposes 'i18n' as a global -->
+        <script src="../../../build/js/live-editor.i18n.js"></script>
+        <!-- Exposes html2canvas as global -->
         <script src="../../../build/js/live-editor.output_webpage_deps.js"></script>
-        <script src="../../../build/js/live-editor.output_webpage.js"></script>
+        <!-- Mocha, sinon, and test JS -->
         <script src="../../../node_modules/mocha/mocha.js"></script>
         <script src="../../../node_modules/expect.js/index.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
```
pamelafox: Removing this copy/pasta sinon.js as well

tests/output/webpage/test_utils.js
pamelafox: Yay! JS and Python had a beautiful ES6 baby!

webpack.config.js
pamelafox: Good point, removing tests from exclude.

pamelafox:
Thank you for the thorough review! So much helpful feedback.

I've addressed the comments and made one significant change as to how the code brings in _ and i18n. The issue before was that test-worker.js brought in pjs-tester.js which relied on those 2, and I had to do various things to make sure test-worker would always bring in _ and i18n locally and webapp.

I decided a cleaner approach is to have a separate webapp config file for the workers. That makes sense, I think, because the externals are different for local build vs worker build vs webapp build.

Specifically:
* For local demos and tests, the bundles currently import _ from lodash and actually include it in the bundles. They also import i18n from the local version and actually include it in the bundles.
* For workers, the bundles need to always have _ and i18n included, as they don't have access to globals. It's okay that test-worker bundle always brings in the local i18n versus the webapp i18n, as it doesn't really utilize it. (We previously mocked it inside test-worker).
* For webapp (in future PR), the bundles treat _ and i18n as externals, and they do *not* include them, since webapp exposes its own versions of them.

Hm. Okay, I must admit, I now realize that the worker bundles could be in the original config, as they *currently* share the same configuration options and externals. They must have been different at some point in my refactoring process. I think the workers still make sense in a separate config, as then it's easy to change some options about them, but it seems like I could move them back if desired.

Time for a brain break :)

webpack.config.js
pamelafox: Ah: the tests in ast_transform_test fail when babel transforms the code, due to the whitespace issues it introduces. I've excluded it for now.

js/output/pjs/pjs-utils.js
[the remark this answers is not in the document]
kevinbarabash: Sorry for using imprecise language... I meant "future PR" when I said "second pass".

package.json
kevinbarabash: webapp uses 6 right now.  I think we'll want to migrate everything to 7 eventually but there were some growing pains moving from 6 to 7 in wonder-blocks that we probably want to avoid at the moment.

tests/output/pjs/iframe.html
kevinbarabash: The changes are fine.  I'm just trying to understand the rationale behind the changes. 🙂

tests/output/webpage/test_utils.js
kevinbarabash: I think we're okay in this case, but we'd definitely want to avoid this in python.

js/output/pjs/pjs-utils.js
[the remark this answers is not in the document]
pamelafox: Ok, noting as a task for future PR.

pamelafox:
Okay, I'll keep workers separate as it does convey the difference in environment, and it might make future differences easier. Merging into my branch (not master!). 

My next PR will likely be the webpackification of the rest of the files (which is definitely a word).

Oh, and re earlier question: loop checking still seems to work fine, in browser and tests.

webpack.config.js
MiguelCastillo: interesting - @kevinbarabash which part is not this not very `webpack-y`?

pamelafox: I think the fact that I'm making a separate bundle for the dependencies, and the dependencies bundle exposes its stuff as globals. The output file makes that more clear, as you can see I include two bundles:
https://github.com/Khan/live-editor/pull/686/files#diff-315326e54ee98fc7fd456045df327d7d

PR #688, opened 2018-10-09
Port editor files to webpack (no more bower!)

pamelafox:

The previous PR ported the output side to webpack, this PR ports the editor side. This PR removes bower entirely and removes build-related tasks from gulp (it leaves test tasks there). 

This branch seems to fully work - all tests are run and are passing, and the demos function. It may be possible to get it working in webapp before moving to React. I'll give that a shot after this PR is merged into my branch.

Notable changes in this PR, besides change to import/export:
* ACE upgraded: there was a change in the format of their deltas, so that required changing some code, mostly in editor-ace and in some tooltips.
* Tomfoolery with images: Getting images to load from url()s in CSS is not so easy. It seems to be working now, thanks to some loaders, but it did involve some URL changes. I imagine that may need tweaking when I build for webapp.
* Module split: I broke TooltipBase into its own file, as it was sharing a file with TooltipEngine. That looks like a large change, but it's really just a file split. I left ImageModal and SoundModal in the same file, as that would have looked even bigger (and there are some actual code changes within them). I'm happy to merge/split as you'd prefer for this diff.
* Tooltip Registration: I added a registerTooltip function to register each tooltip class with the TooltipEngine, as that seems to be a bit better than the old technique (of adding them to TooltipEngine.classes directly).
* Lodash to Underscore: I realized that webapp uses underscore, and there are tiny API differences that matter, so I changed all the imports to underscore, across the editor side and output side.

This is likely making a bigger bundle locally, as there's a lot that goes into the editor- jquery, backbone, bootstrap, jquery plugins, underscore. 
In webapp, the bundle wouldn't be that large (due to more externals) though I don't know how it'd be post the webpack change.
The react change removes all those dependencies except underscore.

No new tests, got old tests running.

### Risks involved

Big change, easy to break things. I'm most wary of CSS/image/plugins related issues.

No blockers for merging into my branch. Can't be merged to master, however, as this isn't necessarily webapp compatible.

Run the demos, run the tests.


js/shared/i18n.js
```diff
@@ -82,7 +82,7 @@ import Jed from "jed";
         // Replace the substitutions with the appropriate option
         for (var i = 1; i < split.length; i += 2) {
             var replaceWith = options[split[i]];
-            split[i] = _.isUndefined(replaceWith) ?
+            split[i] = (replaceWith === undefined) ?
```
pamelafox: This is what i18n has in webapp, seemed better to change it than bring in underscore.

js/ui/tooltip-base.js
```diff
@@ -0,0 +1,171 @@
+import $ from "jquery";
+import Backbone from "backbone";
+Backbone.$ = $;
+
+import "../../css/ui/tooltips.css";
+
```
pamelafox: The file split from tooltip-engine.

js/ui/tooltips/auto-suggest.js
```diff
@@ -43,7 +52,6 @@ TooltipEngine.classes.autoSuggest = TooltipBase.extend({
             this.updateTooltip(lookupParams);
             this.placeOnScreen();
             event.stopPropagation();
-            ScratchpadAutosuggest.enableLiveCompletion(false);
```
pamelafox: LiveCompletion is always off now, so I could remove these.

js/ui/tooltips/color-picker.js
```diff
@@ -108,8 +118,8 @@ TooltipEngine.classes.colorPicker = TooltipBase.extend({
             addSemicolon = true;
         }
 
-        if (event.source && event.source.action === "insertText" &&
-            event.source.text.length === 1 && this.parent.options.type === "ace_pjs" && this.autofill) {
+        if (event.source && event.source.action === "insert" &&
```
pamelafox: ACE upgrade related

js/ui/tooltips/image-picker.js
```diff
@@ -32,7 +39,7 @@ TooltipEngine.classes.imagePicker = TooltipBase.extend({
 
         // TODO(kevinb) extract this into a method on TooltipBase
         if (leadingPadding.length === 0 && path.length === 0 && this.closing.length === 0 &&
-            event.source && event.source.action === "insertText" && event.source.text.length === 1 && this.autofill) {
+            event.source && event.source.action === "insert" && event.source.lines[0].length === 1 && this.autofill) {
```
pamelafox: ACE upgrade related

js/ui/tooltips/number-scrubber.js
```diff
@@ -1,5 +1,10 @@
-// A description of general tooltip flow can be found in tooltip-engine.js
-TooltipEngine.classes.numberScrubber = TooltipBase.extend({
+import $ from "jquery";
+import "webpack-jquery-ui/draggable";
```
pamelafox: I found a library that claimed to make it easy to bring in jquery-ui. It seems to work.

tests/tooltips/colorPicker_test.js
```diff
@@ -1,55 +1,81 @@
+import {getLine, getMockedTooltip, getTooltip, testMockedTooltipDetection, testReplace, typeLine} from "./shared.js";
+import ColorPicker from "../../js/ui/tooltips/color-picker.js";
+
 describe("colorPicker - detection", function() {
-    var mockedColorPicker = getMockedTooltip(tooltipClasses.colorPicker, ["detector", "initialize"]);
+
+    let sandbox;
```
pamelafox: I changed the tooltip tests to use sinon sandbox. It seemed safer now that tests are being bundled together, and its generally a good practice so that you can restore.

tmpl/live-editor.handlebars
```diff
@@ -17,7 +17,7 @@
 
             <div class="scratchpad-canvas-loading">
                 <img src="{{imagesDir}}/spinner-large.gif">
-                <span class="hide-text">{{#_}}Loading...{{/_}}</span>
+                <span class="hide-text">Loading...</span>
```
pamelafox: I removed the {{#_}} calls since they don't actually get handled by webapp (as you recall from my PRs a few weeks ago, and this template in particular gets completely overriden by webapp anyway.

css/ui/tooltips.css
```diff
@@ -288,11 +288,11 @@ html.playing .tooltip.hide-while-playing {
 
 /** Media picker CSS: Used for both an image and audio picker **/
 .mediapicker-modal {
+    background-color: #fff;
```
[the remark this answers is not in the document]
kevinbarabash: Eventually this will be replaced with wonder-blocks-modal, right?

js/shared/i18n.js
kevinbarabash: sgtm... we should really create our own i18n npm package so that it's easy to use here and elsewhere.

js/ui/tooltips/auto-suggest.js
kevinbarabash: Why did we turn off auto-complete?

.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["env"],
```
kevinbarabash: I thought `babel-preset-env` required a `.browserslistrc` to work.  Does it have some sort of default?

.travis.yml
```diff
@@ -3,8 +3,7 @@ sudo: false
 node_js:
   - "8.11.3"
 before_script:
-  - npm install -g bower
-  - bower install
+  - npm install
```
kevinbarabash: yay

css/ui/colorpicker.css
```diff
@@ -3,7 +3,7 @@
 	height: 176px;
 	overflow: hidden;
 	position: absolute;
-	background: url(../images/scratchpads/colorpicker_background.png);
+	background: url(../../images/scratchpads/colorpicker_background.png);
```
kevinbarabash: Why the change in path?

demos/simple/audio.html
```diff
@@ -2,10 +2,6 @@
 <head>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <title>Live Editor Audio Demo</title>
-    <link rel="stylesheet" href="../../build/css/live-editor.core_deps.css"/>
-    <link rel="stylesheet" href="../../build/css/live-editor.audio.css"/>
-    <link rel="stylesheet" href="../../build/css/live-editor.tooltips.css"/>
-    <link rel="stylesheet" href="../../build/css/live-editor.ui.css"/>
```
kevinbarabash: Why are these .css files no longer necessary?

js/shared/i18n.js
kevinbarabash: it looks like this file was just assuming that `_` was a global.  I'm glad we're moving past those dark days.

js/ui/autosuggest-data.js
```diff
@@ -906,3 +908,5 @@ window.ScratchpadAutosuggestData = {
             "keyPressed", "keyReleased", "keyTyped"]
     },
 };
+
+export default AutosuggestData;
```
kevinbarabash: You mentioned that we had turned of auto-suggest... do we still need to the files or are you planning on re-enabling and/or making it configurable in the future?

js/ui/tooltip-base.js
```diff
@@ -0,0 +1,171 @@
+import $ from "jquery";
+import Backbone from "backbone";
+Backbone.$ = $;
+
+import "../../css/ui/tooltips.css";
```
kevinbarabash: I see, you're importing the .css directly in the .js files.  Does this result in the .css appearing in the resulting bundles?

js/ui/tooltip-base.js
kevinbarabash: Cool.  I'm going to assume that there are no changes `TooltipBase`.

js/ui/tooltip-engine.js
```diff
@@ -188,166 +195,8 @@ window.TooltipEngine = Backbone.View.extend({
...
-        this.$el
-            .css({
-                top: $(window).scrollTop() + coords.pageY,
-                left: coords.pageX
-            })
-            .toggle(!(relativePos < 0 || relativePos >= editorHeight));
-    },
+TooltipEngine.registerTooltip = function(name, tooltipClass) {
+    TooltipEngine.classes[name] = tooltipClass;
+};
```
kevinbarabash: You could also do:
```
export registerTooltip = (name, tooltipClass) => TooltipEngine.classes[name] = tooltipClass;
```
and then in files that use `registerTooltip` do:
```
import {registerTooltip} from "./tooltip-engine.js";
```
this is more idiomatic.

tests/tooltips/colorPicker_test.js
kevinbarabash: How does sinon's `sandbox` differ from what we were doing before?

kevinbarabash: Also, these sandbox changes probably would made a good PR on their own.

tmpl/live-editor.handlebars
kevinbarabash: Does this mean we can get rid of the handlebars plugin which was handling `{{#_}}`/`{{/_}}`?

webpack.config.js
```diff
@@ -65,9 +91,33 @@ const config = {
...
+                }
+            },
+            // Needed for loading the Bootstrap font file with WOFF/SVGs
+            {
+                test: /\.(woff|woff2|eot|ttf|svg)$/,
+                loader: 'url-loader?limit=100000'
+            },
         ],
     },
     plugins: [
```
kevinbarabash: Usually people extract the CSS back out of the bundles using either ExtractTextPlugin or MiniCssPlugin.  I'm cool with not doing that.  I'm not sure how keeping the CSS in the bundles affects our ability to override it in webapp.  Maybe that's necessary give that we're using different templates in webapp.

js/ui/tooltips/auto-suggest.js
pamelafox: Good Q. Note that we still pop-up the function arguments hint, but we disabled the auto completion of functions. (Both are handled by ScratchpadAutoSuggest currently)

See comment on https://github.com/Khan/live-editor/blob/a367b6383f9cb1e984935939635e1a80cea5d445/js/ui/tooltip-engine.js#L113 
Here's the commit where I disabled it:
https://github.com/Khan/live-editor/commit/4e028ca5d15d3b062a4a9b38fd4033ddb1a75758

Autosuggest was a hack project, as I recall, so it didn't have time to get well integrated and tested. It'd be better if it could have more time spent on it.

tests/tooltips/colorPicker_test.js
pamelafox: Good call, these changes can go straight to master, I've sent a PR.

pamelafox:
Closing since I'm no longer at KA to make sure this would work within webapp too.

PR #691, opened 2018-10-30
Make row/rows translatable

pamelafox:

This makes it possible to translate "row"/"rows", two strings in the sql-results handlebars template.

No

No, no tests are for translatability.

### Risks involved

Standard danger of short strings: It's possible rows/rows already has a translation in Crowdin that doesn't match our usage well here, so I guess we'll find that out.

No

Open SQL demo locally (http://localhost:8086/demos/simple/sql.html) or /cs/new/sql in webapp

See "row" or "rows" in schema table, depending on number of INSERTs:

CREATE TABLE books (id INTEGER PRIMARY KEY,name TEXT,rating INTEGER);
insert into Books VALUES(1,"Harry Potter", 5); 
insert into Books VALUES(2,"Harry Potter", 5); 

See string show up on CrowdIn a few days after deploy.

### DIFF
--- a/tmpl/sql-results.handlebars
+++ b/tmpl/sql-results.handlebars
@@ -92,9 +92,9 @@ table.sql-schema-table .row-count {
         <table class="sql-schema-table" data-table-name="{{name}}">
         <thead>
         {{#if hasSingleRow}}
-            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} row</span></th>
+            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} {{rowMsg}}</span></th>
         {{else}}
-            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} rows</span></th>
+            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} {{rowsMsg}}</span></th>
         {{/if}}
         </thead>
         <tbody>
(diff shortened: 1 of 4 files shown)


danielhollas:
Thanks, @pamelafox! I checked on Crowdin and neither "row" nor "rows" seems to be there so we should be good in this regard.

davidflanagan:
Sorry the review took so long @pamelafox .  I didn't see the email about it until Daniel commented.

pamelafox:
No worries, Github PRs are easy to miss. I also forgot to assign you, which
is a convention that the WonderBlocks team uses.

Thanks for the review!

pamelafox:
(And by "assign", I mean change the Assignee field as well as the Reviewer
field).

tmpl/sql-results.handlebars
```diff
@@ -92,9 +92,9 @@ table.sql-schema-table .row-count {
         <table class="sql-schema-table" data-table-name="{{name}}">
         <thead>
         {{#if hasSingleRow}}
-            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} row</span></th>
+            <th><a href="javascript:void(0)">{{name}}</a> <span class="row-count">{{rowCount}} {{rowMsg}}</span></th>
```
danielhollas: @pamelafox sorry for commenting late here, but I am not sure this code will handle i18n multiple plurals properly. On the other hand, I am not sure you *can* handle multiple plurals in this framework. Let me know if you need more info.

danielhollas:
Sure, once the strings are on Crowdin I'll be able to make a better
judgement, but it seems to me that this will generate two separate strings,
which is not correct. Caroline has been dealing with the multiple plural
issue a lot so she can probably set you up with sbd. in the know.

tmpl/sql-results.handlebars
danielhollas: Sure, once the strings are on Crowdin I'll be able to make a better
judgement, but it seems to me that this will generate two separate strings,
which is not correct. Caroline has been dealing with the multiple plural
issue a lot so she can probably set you up with sbd. in the know.

danielhollas:
@pamelafox I'm quite confused. Does this PR works for you locally? In my local environment, I don't see the "rows" in the table header. It also doesn't show up on current khan.org

pamelafox:
Huh, strange. I don't see rows in master. I don't know if I was seeing things at the time or if it worked then or what. I'll try your new PR.

PR #692, opened 2018-10-30
Remove unused gutter decorations & underline marker for errors

pamelafox:

We currently immediately render a dashed orange line under an error-ful line as you're typing. This was functionality that was broken that I fixed a few months ago. 

https://khanacademy.slack.com/files/UB9GHCAJF/FDJH16ZV4/screen_shot_2018-10-18_at_7.58.23_am.png

I'm finding the orange line to be very distracting, since it happens immediately and constantly. That isn't in line with our philosophy of giving the user time before we annoy them about errors 
(See: OhNoesOhYes project and this blog post on feedback in online education: https://blog.mrmeyer.com/2016/the-future-of-handwriting-recognition-adaptive-feedback-in-math-education/)

This PR removes the orange line for immediate JS errors. It keeps the orange line for CSS warnings and for Program.assertEqual() failures. It'd be nice for those to be delayed, but they're much rarer and more often signal versus noise.

This PR also removes the code that added gutter decorations for errors, as the code hasn't actually been running for a few years. For a brief period of time, the gutter decoration code would display red error icons, wherever the orange line was displayed. The gutter decorations were buggy (they sometimes displayed for dozens of lines, depending on the error) and the red was jarring.  It appears that we removed them either intentionally or accidentally, and I think that ended up being A Good Thing.  So this PR deletes that not-working-anyway code.

This PR also changes the orange line (for CSS warnings and assertEqual failures) to be a dotted line instead of dashed. The dashed line jiggles too much while you're typing, as it recalculates the dash width. The dotted line doesn't jiggle, as the dots are constant-sized.

Less code, so teeensy tiny bit faster.

There are no tests around this UI.

### Risks involved

Hm. I think one potential beneficial side effect of the orange underline is that it shows up in talk-throughs, so it's clear when the current code has an error. It will be less clear now, since we don't show OhNoes while a talk-through is running. However, given that the orange underline wasn't displaying for a while, it seems fine that it's going to go away again.
(It could be nice to show mini-OhNoes during a talk-through).

No

Go to /cs/new/pjs, type code, don't see orange underline as you type. See thinking Oh Noes if you type an error, click on him to pop up big Oh Noes.
screen shot 2018-10-29 at 7 18 16 pm

Go to /cs/new/webpage, add a CSS rule for a property like "text-crop", see orange line and warning icon.
screen shot 2018-10-29 at 7 21 56 pm

Go to /cs/new/pjs, type Program.assertEqual(2, 4);, see orange line and warning icon.
screen shot 2018-10-29 at 7 20 46 pm

### DIFF
--- a/css/ui/style.css
+++ b/css/ui/style.css
@@ -116,7 +116,7 @@
 }
 
 .scratchpad-ace-editor .ace_problem_line {
-    border-bottom: 1px dashed orange;
+    border-bottom: 1px dotted orange;
     position: absolute;
 }
 
--- a/css/ui/style.css
+++ b/css/ui/style.css
@@ -105,11 +105,6 @@
     z-index: 2;
 }
 
-/* Style overrides for the ace editor */
-.scratchpad-ace-editor .ace_gutter {
-    background: #F5F3E5;
-}
-
 .scratchpad-ace-editor .ace_tooltip {
     max-width: 550px; /* Ensures tooltip isn't cut off */
     white-space: normal; /* Allows multi-lines to wrap */
@@ -122,19 +117,6 @@
 
 .scratchpad-ace-editor .ace_scroller {
     overflow-x: hidden;
-}
-
-.scratchpad-ace-editor .ace_sb {
-    overflow-y: auto;
-}
-
-.scratchpad-ace-editor .ace_active_line {
-    background: rgba(255, 207, 207, 0.9);
-    transition: background 200ms ease-in-out;
-}
-
-.scratchpad-ace-editor .ace_active_line.hilite {
-    background: rgba(255, 100, 100, 1);
 }
 
 .scratchpad-ace-editor .ace_line {
--- a/js/editors/ace/editor-ace.js
+++ b/js/editors/ace/editor-ace.js
@@ -1,6 +1,5 @@
 window.AceEditor = Backbone.View.extend({
     dom: {
-        ACTIVE_LINE: ".ace_active_line",
         TEXT_INPUT: "textarea",
         CONTENT: "div.ace_content"
     },
@@ -300,20 +299,7 @@ window.AceEditor = Backbone.View.extend({
 
     // Set the cursor position on the editor
     setErrorHighlight: function(shouldHighlight) {
-        var self = this;
-
         this.editor.setHighlightActiveLine(shouldHighlight);
-
-        // Delay adding a flash until the active line is shown
-        setTimeout(function() {
-            // Add the hilite flash
-            var line = self.$(self.dom.ACTIVE_LINE).addClass("hilite");
-
-            // And quickly remove it again (to give a nice flash animation)
-            setTimeout(function() {
-                line.removeClass("hilite");
-            }, 100);
-        }, 1);
     },
 
     // Allow for toggling of the editor gutter
--- a/js/live-editor.js
+++ b/js/live-editor.js
@@ -439,12 +439,6 @@ window.LiveEditor = Backbone.View.extend({
             self.record.log("restart");
         });
 
-        // Handle the gutter errors
-        $el.on("click", this.dom.GUTTER_ERROR, function() {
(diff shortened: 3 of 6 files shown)


css/ui/style.css
```diff
@@ -116,7 +116,7 @@
 }
 
 .scratchpad-ace-editor .ace_problem_line {
-    border-bottom: 1px dashed orange;
+    border-bottom: 1px dotted orange;
```
kevinbarabash: mixing it up eh?

css/ui/style.css
```diff
@@ -105,11 +105,6 @@
     z-index: 2;
 }
 
-/* Style overrides for the ace editor */
```
pamelafox: I checked, and out CSS never actually overrides their CSS, so this rule has been void for a while! It'd make the gutter beige, by the way. Doesn't seem needed. :)

css/ui/style.css
```diff
@@ -124,19 +119,6 @@
     overflow-x: hidden;
 }
 
-.scratchpad-ace-editor .ace_sb {
-    overflow-y: auto;
-}
-
-.scratchpad-ace-editor .ace_active_line {
```
pamelafox: "ace_active_line" no longer seems to select anything, and I couldn't find an equivalent class. I couldn't find the string in the ace codebase either.

css/ui/style.css
```diff
@@ -124,19 +119,6 @@
     overflow-x: hidden;
 }
 
-.scratchpad-ace-editor .ace_sb {
```
pamelafox: "ace_sb" also no longer seems to select anything, and I couldn't find it in the ace codebase.

css/ui/style.css
pamelafox: Update: I found the string, it's ace_active-line.

js/editors/ace/editor-ace.js
```diff
@@ -300,20 +299,7 @@ window.AceEditor = Backbone.View.extend({
 
     // Set the cursor position on the editor
     setErrorHighlight: function(shouldHighlight) {
-        var self = this;
-
         this.editor.setHighlightActiveLine(shouldHighlight);
-
-        // Delay adding a flash until the active line is shown
-        setTimeout(function() {
```
pamelafox: At one point, this code flashes a red background on the highlighted line. It no longer does that, since the ACTIVE_LINE class doesn't exist. Now the line is still highlighted gray, thanks to setHighlightActiveLine, but there's no red background.

I can bring back the red (as I know what the new class name is) if desired. However, it doesn't seem needed, given we didn't notice it was gone, plus this ace-dependent code isn't the most stable.

js/live-editor.js
```diff
@@ -1149,11 +1148,10 @@ window.LiveEditor = Backbone.View.extend({
            "ace_problem_line", "text", false);
     },
 
-    removeMarkers: function() {
-        // Remove previously added markers and decorations
+    removeUnderlineMarkers: function() {
```
pamelafox: I renamed this function to be more clear that it removes the UnderlineMarkers

pamelafox:
@kevinbarabash Okay, good idea. I went through and most gutter mentions are about the icons that we still want, for CSS warnings and assertEqual failures. I removed one comment and renamed a function to be more clear.

I then audited the CSS and found a few related rules that are dead ACE CSS, plus some dead JS code.

kevinbarabash:
@pamelafox did you mean to close this?

css/ui/style.css
kevinbarabash: My bad for skimming the commit message.

js/live-editor.js
```diff
@@ -1155,24 +1148,14 @@ window.LiveEditor = Backbone.View.extend({
...
         var session = this.editor.editor.session;
         var markers = session.getMarkers();
-        _.each(markers, function(marker, markerId) {
+        Object.keys(markers).forEach((markerId) => {
```
kevinbarabash: yay

PR #694, opened 2018-11-10
Use Sinon's sandbox for tests

pamelafox:

This change was extracted from https://github.com/Khan/live-editor/pull/688

This PR changes the tooltip tests to use Sinon's sandbox:
https://sinonjs.org/releases/latest/sandbox/
The sandbox makes it easy to clean up fakes (stub(), spy()), which are used throughout the tooltips tests. The tests are less fragile when we clean up fakes after we're done with them.

### Risks involved

Tests pass, seems safe. I've used sandbox in multiple codebases, handy tool.

This can go straight to master. I can then merge master into my webpack PR.

npm run test

### DIFF
--- a/tests/tooltips/colorPicker_test.js
+++ b/tests/tooltips/colorPicker_test.js
@@ -1,55 +1,78 @@
 describe("colorPicker - detection", function() {
-    var mockedColorPicker = getMockedTooltip(tooltipClasses.colorPicker, ["detector", "initialize"]);
+
+    let sandbox;
+    let mockedColorPicker;
+
+    before(function () {
+        sandbox = sinon.sandbox.create();
+        mockedColorPicker = getMockedTooltip(sandbox, tooltipClasses.colorPicker, ["detector", "initialize"])
+    });
+
+    after(function () {
+        sandbox.restore();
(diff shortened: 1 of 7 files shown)


tests/tooltips/colorPicker_test.js
```diff
@@ -1,55 +1,78 @@
...
 
 
 describe("colorPicker - selection (what it replaces)", function() {
-    var mockedColorPicker = getMockedTooltip(tooltipClasses.colorPicker, ["detector", "updateText", "initialize"]);
+    let sandbox;
+    let colorPicker;
+
+    before(function () {
+        sandbox = sinon.sandbox.create();
+        colorPicker = getTooltip(tooltipClasses.colorPicker);
```
pamelafox: This now uses getTooltip() which doesn't set up any internal fakes on the Tooltip class, it just constructs the relevant Tooltip. These tests don't require any internal faked methods.
(This change is also true in the other files, for these types of tests)

kevinbarabash: Nice!  fwiw I like tests that use dependencies as is instead of trying to mock everything out.

PR #696, opened 2018-12-04
Get LiveEditor safely working in Webapp with Webpack

jeresig:
Makes a few different changes to the LiveEditor code to get it working inside Webapp when running both Kake and Webpack. There's another diff forthcoming for the webapp-specific changes.

See also: https://phabricator.khanacademy.org/D50587


external/sqlitejs/sql.js
```diff
@@ -1,6 +1,6 @@
 // This prevents pollution of the global namespace
 var SQL = "Uint8ClampedArray" in window && (function () {
-function f(a){throw a;}var i=void 0,j=!0,k=null,l=!1;function n(){return function(){}}var q,r;r||(r=eval("(function() { try { return Module || {} } catch(e) { return {} } })()"));var aa={},ba;for(ba in r)r.hasOwnProperty(ba)&&(aa[ba]=r[ba]);var ca="object"===typeof process&&"function"===typeof require,da="object"===typeof window,ea="function"===typeof importScripts,fa=!da&&!ca&&!ea;
+function f(a){throw a;}var i=void 0,j=!0,k=null,l=!1;function n(){return function(){}}var q,r;r||(r=eval("(function() { try { return Module || {} } catch(e) { return {} } })()"));var aa={},ba;for(ba in r)r.hasOwnProperty(ba)&&(aa[ba]=r[ba]);var ca=false&&"object"===typeof process&&"function"===typeof require,da="object"===typeof window,ea="function"===typeof importScripts,fa=!da&&!ca&&!ea;
```
jeresig: I disabled this check which was determining if we were "in Node" or not (it's not a very good check, unfortunately).

external/sqlitejs/sql.js
```diff
@@ -318,11 +318,11 @@ function lc(a,b,c){a="string"!==typeof a?x(a):a;try{var d=c?Ub(a,j):Ub(a);C[b>>2
...
-"//");e&&"string"===typeof r.websocket.url&&(g=r.websocket.url);if("ws://"===g||"wss://"===g)g=g+b+":"+c;var h="binary";e&&"string"===typeof r.websocket.subprotocol&&(h=r.websocket.subprotocol);var h=h.replace(/^ +| +$/g,"").split(/ *, */),m=ca?{protocol:h.toString()}:h;d=new (ca?require("ws"):window.WebSocket)(g,m);d.binaryType="arraybuffer"}catch(t){f(new M(J.Lc))}b={$:b,port:c,q:d,Ya:[]};T.M.$b(a,b);T.M.Fd(a,b);2===a.type&&"undefined"!==typeof a.Ba&&b.Ya.push(new Uint8Array([255,255,255,255,112,
+"//");e&&"string"===typeof r.websocket.url&&(g=r.websocket.url);if("ws://"===g||"wss://"===g)g=g+b+":"+c;var h="binary";e&&"string"===typeof r.websocket.subprotocol&&(h=r.websocket.subprotocol);var h=h.replace(/^ +| +$/g,"").split(/ *, */),m=ca?{protocol:h.toString()}:h;d=new (window.WebSocket)(g,m);d.binaryType="arraybuffer"}catch(t){f(new M(J.Lc))}b={$:b,port:c,q:d,Ya:[]};T.M.$b(a,b);T.M.Fd(a,b);2===a.type&&"undefined"!==typeof a.Ba&&b.Ya.push(new Uint8Array([255,255,255,255,112,
```
jeresig: I removed two uses of the "ws" module (which brings in Web Socket support for Node). We don't need them in the browser so it feels safe to remove them.

gulpfile.js
```diff
@@ -34,7 +34,8 @@ gulp.task("templates", function() {
         }))
         .pipe(defineModule("plain"))
         .pipe(declare({
-            namespace: "Handlebars.templates"
+            namespace: "Handlebars.templates",
+            root: "window"
```
jeresig: This makes it so that all the Handlers.templates definitions use "window" instead of "this" (which resolves to undefined in Webpack land).

js/output/webpage/webpage-output.js
```diff
@@ -252,6 +252,7 @@ window.WebpageOutput = Backbone.View.extend({
         _.extend(this.tester.testContext, {
             $doc: $(this.frameDoc),
             docSP: this.slowparseResults.document,
+            $docSP: $(this.slowparseResults.document),
```
jeresig: I'm not sure if this just changed recently or what - but all of my webpage challenges were failing locally because this was missing, so I added it in?

pamelafox: This did change recently (as I removed jQuery from webpage-output.js), and there were a few webpage challenges that I updated (like Write a Poem). Does your localhost have the latest challenges?

external/sqlitejs/sql.js
pamelafox: Hm, in my webpack PR, I did require("sql.js") (straight from node_modules), does this mean we wouldn't want to require sql.js because it brings in too much?

jeresig: I'm not sure! I wonder if the NPM version of this is smarter. We're using a really old version so I think it wasn't built in a particularly effective way.

js/output/webpage/webpage-output.js
jeresig: Ahh - I must not have the latest challenges locally then. I could remove this or keep it in - what would you prefer? (Keeping it in feels safer in the near term, at least until we fully transition away from jQuery.)

pamelafox: It's been out of the live codebase for a few months, so I'd vote for keeping it removed. The challenges should work on ZND.

jeresig:
@pamelafox I think I resolved the one issue you had. Let me know if you have any other concerns (also: @kevinbarabash!)

PR #699, opened 2019-01-14
Fix 1 of 2 BabyHint banned properties bypass exploits

aepic999:

I am fixing an exploit where a constructor chain from any Object can be used to circumvent security features. I created the ticket 228409 about this and another similar issue.
I have not tested, but it should be very little (maybe a few ms at most.)

No tests have been added. I didn't see a test file for BabyHint in the test folder.

### Risks involved

If someone is attempting to define a property named "constructor" on any object, it may break some programs.

No new dependencies are required.

Attempt to use a constructor to create a Function object, like this:

```[].constructor.constructor('console.log("Hi!");')();```
If an error appears through Oh Noes and nothing appears in the console, the change has worked.

### DIFF
--- a/js/output/pjs/babyhint.js
+++ b/js/output/pjs/babyhint.js
@@ -139,7 +139,8 @@ var BabyHint = {
         location: true,
         document: true,
         ownerDocument: true,
-        createElement: true
+        createElement: true,
+        constructor: true // Can be used to bypass some circumventions
     },
 
     variables: [],
--- a/tests/output/pjs/jshint_test.js
+++ b/tests/output/pjs/jshint_test.js
@@ -108,6 +108,13 @@ describe("Scratchpad Output - BabyHint checks", function() {
         babyhint: true,
         code: "rect(100, 200, 300);"
     });
+    
+    assertTest({   
+        title: "Banned properties test",
+        reason: "constructor is a reserved word.",
+        babyhint: true,
+        code: "[].constructor.constructor('return 1');"
+    });
 });
 
 // Syntax errors - not controlled by JSHint options.
@@ -147,6 +154,7 @@ describe("Scratchpad Output - JSHint syntax errors", function() {
         code: "10"
     });
 
+    
     assertTest({
         title: "Not closing a string",
         reason: "Unclosed string! Make sure you end your string with a quote.",

### CODE REVIEW

**Comment by pamelafox** [comment]
Thanks for the PR! This file contains both BabyHint and JSHint tests (it is poorly named): https://github.com/Khan/live-editor/blob/master/tests/output/pjs/jshint_test.js
Can you add a test there?

Also, can you brainstorm valid situations where people may currently be using constructor? I can grep the database to see how frequent they are and see whether this may break existing programs.

**Comment by aepic999** [comment]
Thank you. I just committed the test, and added a possible risk.

**Comment by pamelafox** [comment]
Ok, I did a quick look for programs that may be affected.

Here's a program that references a constructor property:
https://www.khanacademy.org/computer-programming/ellipsex-y-w-h/6636169099378688

Here's another program that calls Object.constructor():
https://www.khanacademy.org/computer-programming/ellipsex-y-w-h/4691443639025664

Please look into those two examples and comment on whether they would be affected. (There are quite a few others, but many are probably similar).

**Comment by aepic999** [comment]
6636... appears to use the constructor in terms of it's own object. I could probably add a second test that checks if it's not only a constructor property, but if the type is a Function (which is the real circumvention here).
4691... looks like it's using the constructor abuse exploit to create arrays. This would clearly be affected, even with above circumvention described for 6636... since it is directly using the exploit.

**Comment by aepic999** [comment]
Also one interesting thing appeared to me looking through 6636...
```var createCanvas = function(){
    return this[["externals"]][["canvas"]][["ownerDocument"]][["createElement"]]("canvas");
};```
It appears to be somehow bypassing the createElement restriction and access to document by accessing externals. This may be a problem. Maybe I'll add `ownerDocument` as a banned keyword.

**Comment by aepic999** [comment]
Nevermind, that's already a thing. I'm curious as to how this was pulled off.

**Comment by aepic999** [comment]
Trying `var a = this[["externals"]];` gets me an error: `This program uses capabilities we've turned off for security reasons. Khan Academy prohibits showing external images, playing external sounds, or displaying pop-ups.`
I'm wondering how this program managed to accomplish a ton of things that shouldn't work.

**Comment by aepic999** [comment]
So, apparently, you can return values, even if they can't be accessed normally.
`var a=this[["externals"]]` doesn't work, but
```
var b=function() {
    return this[["externals"]];
};
var a=b();
```
works just fine.

**Comment by aepic999** [comment]
Turns out that the anomalies of 6636... caused yet another sandbox bypass:
```
var injectEval=function(code){
    var x=this[["externals"]][["canvas"]][["ownerDocument"]][["createElement"]]('script');
    x.type='text/javascript';
    x.text=code;
    this[["externals"]][["canvas"]][["ownerDocument"]].body.appendChild(x);
};
```

**Comment by pamelafox** [comment]
Hm. Our <iframe> doesn't set allow-popups on ProcessingJS programs, so window.open() should not work. What browser are you using that you are seeing pop-ups for the pwned program?

**Comment by aepic999** [comment]
Firefox.

**Comment by aepic999** [comment]
Interesting. Just tried Pwned v2 in another Firefox environment and it doesn't work. Test environment this time was Firefox 64.0.2. Console gives me `InvalidAccessError: A parameter or an operation is not supported by the underlying object`

**Comment by aepic999** [comment]
Since the sandbox interfaces with the main page, if we were to compromise something in the main page using the sandbox, we could execute arbitrary scripts in the context of the main page and possibly disclose information that shouldn't be accessible by the sandbox. I'll dig through the code for possible problematic spots that could lead to a compromise of the main page.

**Comment by pamelafox** [comment]
Hm, Firefox should fully honor iframe sandbox="allow-popups" (see table at bottom of https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe). I wonder if your Firefox has different settings.

Generally, the big concern is whether programs can actually access user data outside the sandbox. So far, it does not seem to be the case for this vector. If you do manage to get outside the sandbox in a standard browser, please let us know. (It may be better to report on ZenDesk in that case, for safer disclosure). Thanks!

**Comment by aepic999** [comment]
More bypasses:
```
var b=Object.hasOwnProperty;
var reenable = {
        "document": true,
        "ownerDocument": true,
        "createElement": true,
        "constructor": true,
        "location": true
};
var x=function(property){
    if (reenable[property]){
        return false;
    } else {
        return this[property]!==undefined;
    }
};
Object.prototype.hasOwnProperty = x;
```

aepic999:
I'm limited as to what I can test right now, I'll start sifting through code when I get home, lol

pamelafox:
Thanks for investigating! Since we've determined that this particular change would affect a number of existing programs and that our domain sandboxing prevents the exploit in our supported browsers, I'm going to close this PR for now. Please let us know if you find you're able to reach outside the iframe sandbox.

PR #700, opened 2019-02-11
Add logFullScreen option 

pamelafox:

This PR adds a logFullScreen option to LiveEditor, which then gets passed to LiveEditorOutput and then to PJSOutput. It ultimately gets used to trigger a full-screen log after the Processing sketch loads. 

(See my related PR in processing-js repo: https://github.com/Khan/processing-js/pull/12)

This PR is 99% option forwarding.

No, and will not be used by majority of programs anyway.

No. 

### Risks involved

There's the usual risk in adding a new option in that it adds slightly more complexity and maintenance burden.

Yes, external/processing-js PR

Locally, I edited LiveEditor demos/simple/index.html to set logFullScreen to true and verified that worked. I also checked I could trigger the processingJS change by simply typing println("LOGFULLSCREEN").

There's no control for it in webapp yet, so only the latter test would work there.

### DIFF
--- a/js/output/pjs/pjs-output.js
+++ b/js/output/pjs/pjs-output.js
@@ -166,6 +166,9 @@ window.PJSOutput = Backbone.View.extend({
     build: function(canvas, enableLoopProtect, loopProtectTimeouts) {
         this.processing = new Processing(canvas, (instance) => {
             instance.draw = this.DUMMY;
+            if (this.options.logFullScreen) {
+              instance.println("FULLSCREENLOG");
+            }
         });
 
         // The reason why we're passing the whole "output" object instead of
@@ -225,6 +228,10 @@ window.PJSOutput = Backbone.View.extend({
 
             // Restart execution
             this.output.restart();
+        }
+        if (this.processing) {
+            // find log
+
         }
     },
 
--- a/js/output/pjs/pjs-output.js
+++ b/js/output/pjs/pjs-output.js
@@ -229,10 +229,6 @@ window.PJSOutput = Backbone.View.extend({
             // Restart execution
             this.output.restart();
         }
-        if (this.processing) {
-            // find log
-
-        }
     },
 
     messageHandlers: {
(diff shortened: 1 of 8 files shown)


js/output/pjs/pjs-output.js
```diff
@@ -166,6 +166,9 @@ window.PJSOutput = Backbone.View.extend({
     build: function(canvas, enableLoopProtect, loopProtectTimeouts) {
         this.processing = new Processing(canvas, (instance) => {
             instance.draw = this.DUMMY;
+            if (this.options.logFullScreen) {
+              instance.println("FULLSCREENLOG");
+            }
```
kevinbarabash: It looks like this may have been used for testing purposes.  Please remove before merging.

kevinbarabash:
I tried re-running the travis build but that did not help. :(

kevinbarabash:
@pamelafox I tried click on https://github.com/Khan/processing-js/compare/66bec3a3ae88262fcb8e420f7fa581b46f91a052...5d411aa4081fbae04fda71dfb41ff0dc9ff6e7d2 to see the changes in `processing-js` but there's nothing there.  Is it possible that the commit that this PR is referencing doesn't exist in `Khan/processing-js`?

pamelafox:
@kevinbarabash I updated the submodule to point at the pushed PR (https://github.com/Khan/processing-js/pull/12/files). I think that's also why Travis kept failing, it just passed.

js/output/pjs/pjs-output.js
pamelafox: This is on purpose - this is what triggers the full-screen mode in the ProcessingJS logger. I describe more in the PR: https://github.com/Khan/processing-js/pull/12/files
I can play around with other approaches as desired, however.

pamelafox:
Screenshot from live-editor simple demo:
screen shot 2019-02-12 at 11 19 00 pm

PR #701, opened 2019-02-13
Fixing visible logic for LoopProtect

pamelafox:

In PR #670, I replaced the visibly library with the native visibilitychange event. Sadly, when I did that, I reversed the logic for setting visible after a visibilitychange. This meant that if you changed tabs after starting to code in the editor and _then_ introduced an infinite loop, the tab would crash. Sad! :( :(
(Hopefully most students reloaded in a new tab and were okay, so perhaps that's why we didn't see a huge number of reports caused by this regression).

This PR adds a single character to fix the bug.

No, there are no tests that test the visibility change. It may be possible to mock it but that'd add a lot of time to this fix.

### Risks involved

This should bring it back to its previous functional state, so it should be low risk. 

Open the local demo. Paste this code:

for (var i = 0; i < 10;) {
   
}

See that the loop error pops up.

Now delete i < 10; in that code.

Switch to any other tab, switch back.

Paste in the code again. 

See that the loop error pops up after a second of thinking.

Once deployed into webapp, a great place to test is Loopy Ruler, as that's where students have run into it.

screen shot 2019-02-13 at 11 44 56 am

### DIFF
--- a/build/js/live-editor.output_pjs_deps.js
+++ b/build/js/live-editor.output_pjs_deps.js
@@ -87486,6 +87486,7 @@ window.LoopProtector.prototype = {
                         _this2.callback(_error);
                         throw _error;
                     }
+
                     // Determine which of KAInfiniteLoopProtect's callsites has
                     // the most calls.
                     var max = 0; // current max count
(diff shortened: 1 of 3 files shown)


kevinbarabash:
Thanks for fixing this.

PR #703, opened 2019-02-15
Fix #528

### LINKED ISSUE #528: Lint error makes for a comatose program.
See the program at https://www.khanacademy.org/computer-programming/wat5/5754244333961216 

A syntax error (?) is not flagged by Yoko Oh Noes and the program appears comatose (since it is not run).

EthanLuisMcDonough:
Prevents blank slots in arrays from silently crashing the program (Fixes #528).

No.

Yes. See `output_test.js`.

### Risks involved
None that I know of.

No.

I added tests.  Opening up the demo and writing something like `var h = [,,]; println(10);` also works.


js/output/shared/ast-walker.js
```diff
@@ -23,7 +23,10 @@ window.walkAST = function(node, path, visitors) {
             if (Array.isArray(node[prop])) {
                 let i = 0;
                 while (i < node[prop].length) {
-                    let child = node[prop][i];
+                    // Esprima represents empty slots as null, so set child to
+                    // a dummy node if the initial node is falsy in order to prevent
+                    // a silent crash
```
pamelafox: Are there other situations where the node would be false-y besides the empty slot? I'm wondering if it's helpful to have a test case for undefined/null/0.

pamelafox:
Oh, and a clarification on the PR template. This question "How can we verify that this change works?" is a question about how to manually verify the change once deployed. Having a manual verification in addition to automated tests helps us verify our deploys and make sure that the code truly made it into the build. You did add a a manual test under this PR's question, so that's what I'll use for the deploy.

js/output/shared/ast-walker.js
EthanLuisMcDonough: If [modern esprima's type definitions](https://github.com/jquery/esprima/blob/master/src/nodes.ts) are anything to go off of, `null` should be the only falsey value that shows up in node array-properties.  I don't think they'd change their AST structure significantly throughout versions.

pamelafox: Okay, thanks, that link clarifies.

pamelafox:
This change was deployed to khanacademy.org yesterday, by the way. Thanks for it!

PR #705, opened 2019-02-26
Fix #400

### LINKED ISSUE #400: calls to the .toString() on functions should output code that is the same as the user code

EthanLuisMcDonough:
Resolves #400. The `toString` method on functions now returns the learner's original code rather than the AST-transformed code.

The AST transformation might be a bit slower, but I don't think this change should impact the project's overall speed that much at all.

Yes.

### Risks involved
None that I know of.

No.

Testing this change really only involves displaying a function as a string and seeing if you get the original.  Anything from `println(function() { /* ahakdhaksdjh*/ var h = 20;  });` to 
```
Program.assertEqual(function() {
    var h = 20;
    for (;;) {
        h++;
    }
}.toString(), "function() {\n    var h = 20;\n    for (;;) {\n        h++;\n    }\n}");
```
works.


pamelafox:
Sorry for the delay in reviewing this, Ethan! This is a larger PR, so I won't have time to look at it until after April 19th, as that's my deadline for releasing the KA AP CS P course. 
Just wanted to let you know that I definitely saw it, appreciated it, and have it in my inbox as a TODO.

PR #717, opened 2019-10-04
Make SQL 'ambiguous column name' error translatable

danielhollas:

The errors thrown by SQLite are parsed individually and made more user-friendly and marked for translations. This one particular error -- `ambiguous column name` -- is not handled, which means the corresponding string cannot be translated into other langs on Khan. This commit fixes that. Also, I'll try to make the text more helpful, but I am not an SQL expert so this needs to be reviewed.

Tracking JIRA ticket: [IC-527](https://khanacademy.atlassian.net/browse/IC-527)

Minimal. 

Not sure whether there are existing tests for this functionality.

### Risks involved

I simply copied the existing code and only changed a couple of strings and variable names, hopefully, it should be safe.

Nope.

1. run: `python -m SimpleHTTPServer`
2. Go to: http://0.0.0.0:8000/demos/simple/sql.html
3. This SQL should trigger the error, you can verify that the new text appears instead of the old one.
```sql
CREATE TABLE student_grades (id INTEGER, grade INTEGER);
INSERT INTO students VALUES (1, "Brian", 3);
INSERT INTO student_grades VALUES (1, 95);
SELECT id FROM students INNER JOIN student_grades on students.id = student_grades.id;
```

### DIFF
--- a/js/output/sql/sql-output.js
+++ b/js/output/sql/sql-output.js
@@ -96,6 +96,13 @@ window.SQLOutput = Backbone.View.extend({
             errorMessage = i18n._("You have multiple columns named \"%(colName)s\" - " +
                 "column names must be unique.", {colName});
         }
+        const ambColStr = "ambiguous column name:";
+        const ambColError = sqliteError.indexOf(ambColStr) > -1;
+        if (ambColError) {
+            const colName = errorMessage.split(ambColStr)[1].trim();
+            errorMessage = i18n._("Ambiguous column name \"%(colName)s\".",
+                {colName});
+        }
         const unknownColStr = "no such column:";
         const unknownColError = sqliteError.indexOf(unknownColStr) > -1;
         if (unknownColError) {
--- a/build/js/live-editor.output_sql.js
+++ b/build/js/live-editor.output_sql.js
@@ -838,6 +838,12 @@ window.SQLOutput = Backbone.View.extend({
             var colName = errorMessage.split(dupColStr)[1].trim();
             errorMessage = i18n._("You have multiple columns named \"%(colName)s\" - " + "column names must be unique.", { colName: colName });
         }
+        var ambColStr = "ambiguous column name:";
+        var ambColError = sqliteError.indexOf(ambColStr) > -1;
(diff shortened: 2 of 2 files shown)


js/output/sql/sql-output.js
```diff
@@ -96,6 +96,13 @@ window.SQLOutput = Backbone.View.extend({
             errorMessage = i18n._("You have multiple columns named \"%(colName)s\" - " +
                 "column names must be unique.", {colName});
         }
+        const ambColStr = "ambiguous column name:";
+        const ambColError = sqliteError.indexOf(ambColStr) > -1;
+        if (ambColError) {
+            const colName = errorMessage.split(ambColStr)[1].trim();
+            errorMessage = i18n._("Ambiguous column name \"%(colName)s\".",
```
danielhollas: Maybe we could help learners a bit more here and something like this:

`When joining two tables which have columns with an identical name, prefix the column name with the table name like this table1.\"%(colName)s\"`

aasmundeldhuset: Good idea! I suggest a rewording: `This means that two or more of the tables that you are joining contain a column with that name. When you want to use that column in your query, you need to say which table you are talking about. Example: firstTable.\"%(colName)s\"` Too bad that we can't get the actual table names from the query...

pamelafox: Nice user-friendly wording. It might be a bit too long for Yoko OhNoes, so here's a shorter version of that if needed:
"Multiple tables that you're joining contain a column with that name. To use that column in your query, specify the table of the column. Example: firstTable.\"%(colName)s\""

danielhollas: @aasmundeldhuset okay, SGTM.
Since this is already live, I'll add this change to the current PR #724 if that's okay. I will make it a separate string so that we do not break existing translations.

danielhollas: Thank you @aasmundeldhuset  and @pamelafox. In the end, I created a separate PR #717 and added a missing test.

PR #719, opened 2019-11-02
Making loop messages translatable

pamelafox:

Addresses https://khanacademy.atlassian.net/browse/CTI-307: it is difficult to translate the loop checker message into gendered languages like Spanish where function and loop have different genders. 

This makes the message for each construct a separate sentence, for ultimate translatability across languages. Please do let me know if there's a slicker way to do this with i18n._().

No.

No, the loop protector tests don't currently test the message output itself.

### Risks involved

Weird spacing? I wasn't sure the best way to ensure a space between the first sentence and the second. 
Also, we will likely lose any current translations.

No.

Open demos/simple.html, write an infinite loop, see message pop up. Do same on cs/new/pjs once this is in webapp.

### DIFF
--- a/build/js/live-editor.ui.js
+++ b/build/js/live-editor.ui.js
@@ -2184,11 +2184,11 @@ window.LiveEditor = Backbone.View.extend({
 
     cleanErrors: function cleanErrors(errors) {
         var loopProtectMessages = {
-            "WhileStatement": i18n._("<code>while</code> loop"),
-            "DoWhileStatement": i18n._("<code>do-while</code> loop"),
-            "ForStatement": i18n._("<code>for</code> loop"),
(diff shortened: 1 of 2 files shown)


danielhollas:
@pamelafox I still don't see the new strings on Crowdin, only the [old string](https://crowdin.com/translate/khanacademy/26335/enus-cs#3279365). Has this been deployed?

pamelafox:
@danielhollas Yep, it was deployed, I found the strings. Here's while loop, for example: https://crowdin.com/translate/khanacademy/26342/enus-cs#6543164

danielhollas:
Thank you! Sorry for a false alarm, I probably had a filter in Crowdin search that I forgot about.

PR #724, opened 2020-02-22
Make SQL error translatable

danielhollas:

Fixing untranslatable SQL error string coming originally from SQLite. (same issue as in #717).

`table XX has X columns but Y values were supplied`

JIRA: [IC-611](https://khanacademy.atlassian.net/browse/IC-611)

This PR enhances the English output as well since it was not properly pluralized in SQLite.

No.

Yes. I've also added a test covering #717.

### Risks involved

I hope that the approach for i18nizing this sentence is correct and will generate correct strings on Crowdin. :crossed_fingers: 

No.

1. Play with the number of values in the following INSERT commands, verify the error is printed and pluralized correctly for different number of columns and values.

```sql
CREATE TABLE student_grades (id INTEGER);

INSERT INTO student_grades VALUES (1);
INSERT INTO students VALUES (1, "sal", 1);
```

2. When this lands, verify that the new Crowdin strings are properly pluralized.

### DIFF
--- a/js/output/sql/sql-output.js
+++ b/js/output/sql/sql-output.js
@@ -122,6 +122,39 @@ window.SQLOutput = Backbone.View.extend({
             errorMessage = i18n._("There's a syntax error near %(nearThing)s.",
                 {nearThing: nearPhrase.substr(5)});
         }
+        const colValueMismatchReg = /has (\d+) columns but (\d+) values were supplied/;
+        const colMismatchError = sqliteError.match(colValueMismatchReg);
+        if (colMismatchError) {
+          const tableName = sqliteError.split(" ")[1];
+          const numCols = colMismatchError[1];
+          const numVals = colMismatchError[2];
+
+          // I18N: The first part of "Oh noes" error in the SQL course.
+          // I18N: "Table %(tableName)s has %(num)s columns"
+          // I18N: "but %(num)s values were supplied."
+          // I18N: You can reorder the parts in string "%(columns)s %(values)s"
+          const tableHasColumns = i18n.ngettext(
+              "Table \"%(tableName)s\" has %(num)s column",
+              "Table \"%(tableName)s\" has %(num)s columns",
+              numCols, {tableName: tableName});
+
+          // I18N: The second part of "Oh noes" error in the SQL course.
+          // I18N: "Table %(tableName)s has %(num)s columns"
+          // I18N: "but %(num)s values were supplied."
+          // I18N: You can reorder the parts in string "%(columns)s %(values)s"
+          const valuesSupplied = i18n.ngettext(
+              "but %(num)s value was supplied.",
+              "but %(num)s values were supplied.",
+              numVals);
+
+          // I18N: Oh noes error in the SQL course.
+          // I18N: %(columns)s = "Table %(tableName)s has %(num)s columns"
+          // I18N: %(values)s  = "but %(num)s values were supplied."
+          // I18N: The sentences are translated in two different Crowdin strings,
+          // I18N: but can be reordered here
+          errorMessage = i18n._("%(columns)s %(values)s",
+              {columns: tableHasColumns, values: valuesSupplied});
+        }
 
         // Now that we've translated the base error messages,
         // we add on additional helper messages for common mistakes
--- a/tests/output/sql/output_test.js
+++ b/tests/output/sql/output_test.js
@@ -56,6 +56,14 @@ describe("Linting", function() {
     failingTest("Inserting rows mismatching quote",
         "INSERT characters ('Summer);",
         ["There's a syntax error near \"characters\"."]);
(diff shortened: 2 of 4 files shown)


danielhollas:
@aasmundeldhuset would you mind checking that I took the correct approach for i18nizing a sentence with two numerical variables? Thanks!

js/output/sql/sql-output.js
```diff
@@ -122,6 +122,26 @@ window.SQLOutput = Backbone.View.extend({
...
+              "Table \"%(tableName)s\" has 1 column",
+              "Table \"%(tableName)s\" has %(num)s columns",
+              numCols, {tableName: tabName});
+
+          const valuesSupplied = i18n.ngettext(
+              "but 1 value was supplied.",
+              "but %(num)s values were supplied.",
+              numVals);
+
+          errorMessage = i18n._("%(columns)s %(values)s",
```
danielhollas: Is there a way to add context for this strings so translators are aware what this referes to?

pamelafox: This is how we do it in the main codebase:
// I18N: "bold" as in bolded text that's thicker for emphasis.
bold: i18n._("bold"),

I think it might work here? It's worth a shot.

tests/output/webpage/output_test.js
```diff
@@ -266,6 +266,7 @@ describe("Linting", function() {
     if (!isFirefox()) {
         // An exception occurs in slowparse when parsing this HTML on
         // Chrome, Safari, and phantomjs.
+        // NOTE(danielhollas): This test is failing in my local machine
```
pamelafox: You could change that note "This test seems to be failing on local machines" since it's also true for my machine. I would assume it's due to a difference in the Chrome browsers but I haven't dug into it.

tests/output/sql/output_test.js
```diff
@@ -116,6 +124,13 @@ describe("Linting", function() {
         "INSERT INTO character_episodes VALUES (2, 'Rick');" +
         "SELECT * FROM characters LEFT OUTER JOIN character_episodes ON " +
         "id = characterId");
+    failingTest("Ambiguous column name when joining",
+	"CREATE TABLE students (id INTEGER, name TEXT);" +
+	"CREATE TABLE student_grades (id INTEGER, grade INTEGER);" +
+	"INSERT INTO students VALUES (1, \"Brian\");" +
+	"INSERT INTO student_grades VALUES (1, 95);" +
+	"SELECT id FROM students INNER JOIN student_grades on students.id = student_grades.id;",
+	["Ambiguous column name \"id\"."]);
```
pamelafox: How does this test relate to the change?

danielhollas: I did not know how to add tests before so I added an additional test covering previous PR #717.

js/output/sql/sql-output.js
aasmundeldhuset: Yes, `// I18N: Explanation here` works in JS too. In `ngettext`, the third parameter is used to select the plurality at runtime, and also automatically becomes the named parameter `%(num)s`, so it's correct that the parameter dict in the first call only needs to contain `tableName` and that the second call doesn't need a parameter dict. However, 1 must not be hardcoded in the singular string, because then the linter won't let the translation contain `%(num)s`, and some languages use the "singular case" for other numbers than 1.

danielhollas: @aasmundeldhuset thank you for confirming and for spotting an issue. What a silly mistake for me to make, since I am well aware of this problem. Not sure what I was thinking. :thinking: Will fix and add a comment soon.

danielhollas:
@pamelafox @aasmundeldhuset thanks for the review! I've rebased on master, fixed the pluralization and added comments for translators.

js/output/sql/sql-output.js
```diff
@@ -122,6 +122,39 @@ window.SQLOutput = Backbone.View.extend({
             errorMessage = i18n._("There's a syntax error near %(nearThing)s.",
                 {nearThing: nearPhrase.substr(5)});
         }
+        const colValueMismatchReg = /has (\d+) columns but (\d+) values were supplied/;
+        const colMismatchError = sqliteError.match(colValueMismatchReg);
+        if (colMismatchError) {
+          const tabName = sqliteError.split(" ")[1];
```
pamelafox, line 128: Nit: I'd find this more readable if it was tableName, as I pronounced "tab" like a browser tab and got confused as to what it was talking about.
All of these rely on array indices existing, but I believe they're all guaranteed based on earlier matching, right?

danielhollas: Done, thank you.

tests/output/sql/output_test.js
danielhollas: I moved the test to #729 since it would have to be updated anyways.

PR #725, opened 2020-03-17
Rewrite function declarations

pamelafox:

This rewrites function declarations into variable declarations. It does the rewriting when all the other rewriting is done, which happens after linting and tests are run.

The goal is that I can write examples for AP CSP that use function declaration syntax, as that is more compatible with how Code.org students are learning functions. That makes it easier for Code.org students/teachers to use the KA materials for remote learning.

It's one more check while walking the AST. Should be negligible?

Yes, I added several in ast-transform. Happy to add any more edge cases you can think of.

### Risks involved

I think the AP CSP examples would be fine/not risky, I can put this on a ZND to see.
However, If other people start using this syntax, they may run into unforeseen issues with it: interaction with real-time update, interaction with challenge graders, etc. If they do run into issues, I disabled the babyhint so they would not see any message that'd indicate why something went awry.

Do you think this could put programs at risk somehow that aren't using function declarations? (Are there nodes with type=FunctionDeclaration besides what I'm testing?)

webapp would need to be updated upstream

Try this program, see rect and ellipse:

```
function dry(x, y) {
    function innerFunc(x, y) {
        ellipse(x, y, 16, 13);
    }
    rect(x, y, 167, 137);
    innerFunc(102, 88);
}

dry(120, 95);

```

### DIFF
--- a/js/output/pjs/babyhint.js
+++ b/js/output/pjs/babyhint.js
@@ -218,8 +218,6 @@ var BabyHint = {
 
             // Checks could detect new errors, thus must run on every line
             errors = errors
-                // check for incorrect function declarations
-                .concat(BabyHint.checkFunctionDecl(line, lineNumber))
                 // we don't allow ending lines with "="
                 .concat(BabyHint.checkTrailingEquals(line, lineNumber))
                 // check for correct number of parameters
--- a/js/output/pjs/pjs-ast-transforms.js
+++ b/js/output/pjs/pjs-ast-transforms.js
@@ -318,3 +318,29 @@ ASTTransforms.rewriteNewExpressions = function(envName) {
         }
     }
 };
+
+
+
+ASTTransforms.rewriteFunctionDeclarations = {
+    leave(node, path) {
+        if (node.type === "FunctionDeclaration") {
+            var decl = {
+                type: "VariableDeclarator",
+                id: {
+                    type: "Identifier",
+                    name: node.id.name
+                },
+                init: {
+                    type: "FunctionExpression",
+                    id: null,
+                    params: node.params,
+                    body: node.body,
+                    generator: node.generator,
+                    expression: node.expression,
+                    async: node.async
+                }
+            };
+            return b.VariableDeclaration([decl], "var");
+        }
+    }
+};
--- a/tests/output/webpage/output_test.js
+++ b/tests/output/webpage/output_test.js
@@ -263,24 +263,11 @@ describe("Linting", function() {
         ]
     );
 
-    if (!isFirefox()) {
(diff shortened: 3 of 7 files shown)


tests/output/webpage/output_test.js
```diff
@@ -263,24 +263,11 @@ describe("Linting", function() {
...
-
-    if (isFirefox()) {
-        // slowparse succeeds when parsing this HTML on Firefox.
-        failingTest("Not so fatal slowparse error detected (Firefox)",
-            "<li><a href='</li><img src='https://www.kasandbox.org'>", [
-                {row: 0, column: 7, lint: {type: "INVALID_ATTR_NAME"}}
-            ]
-        );
-    }
+    failingTest("Slowparse error detected",
```
pamelafox: It used to be that Chrome and Firefox behaved differently, but they now appear to act the same (at least locally, let's see what TravisCI says)

danielhollas:
FWIW that "slow parse error" test was also failing for me locally when I was working on #724. Glad to know I am not the only one. :slightly_smiling_face:

js/output/pjs/pjs-ast-transforms.js
```diff
@@ -318,3 +318,29 @@ ASTTransforms.rewriteNewExpressions = function(envName) {
...
+                    name: node.id.name
+                },
+                init: {
+                    type: "FunctionExpression",
+                    id: null,
+                    params: node.params,
+                    body: node.body,
+                    generator: node.generator,
+                    expression: node.expression,
+                    async: node.async
```
kevinbarabash: just in case 😉

js/output/pjs/babyhint.js
```diff
@@ -218,8 +218,6 @@ var BabyHint = {
 
             // Checks could detect new errors, thus must run on every line
             errors = errors
-                // check for incorrect function declarations
-                .concat(BabyHint.checkFunctionDecl(line, lineNumber))
```
kevinbarabash: It's nice that we don't have to disable all of BabyHint.

pamelafox:
@danielhollas good to know! Must be some environment mismatch.

@kevinbarabash Thanks for the review! I deployed to a ZND and things are looking good there so far:  https://static-znd-200317-pamela-p.khanacademy.org/computer-programming/new/pjs

danielhollas:
@pamelafox not sure if this is of any concerns for AP CSP student, but potentially confusing thing is that function hoisting does not work in this approach:

image

pamelafox:
@danielhollas I'm fine with that, as I think it's better practice to define your functions first. (More cross-language friendly). But thank you for bringing it up, it's useful to track the effects of this change/approach.

PR #729, opened 2020-04-05
Add more learner-friendly explanation and missing test for ambiguous column error

danielhollas:

This PR is a followup on #717, adding a more detailed explanation to a rather cryptic ambiguous column error. I also added a missing test.

No.

I added a missing test.

### Risks involved
To avoid deleting existing translations, I added the new message as a separate string.
It looks a bit ugly in the code but is translator-friendly (I checked that more than 8 lang translated the error already). But LMK if you want to change it, it's not a deal-breaker.

No.

The following code triggers the error at http://localhost:8000/demos/simple/sql.html
```sql
CREATE TABLE student_grades (id INTEGER, grade INTEGER);
INSERT INTO students VALUES (1, "Brian", 3);
INSERT INTO student_grades VALUES (1, 95);
SELECT id FROM students INNER JOIN student_grades on students.id = student_grades.id;
```
image

### DIFF
--- a/js/output/sql/sql-output.js
+++ b/js/output/sql/sql-output.js
@@ -102,6 +102,14 @@ window.SQLOutput = Backbone.View.extend({
             const colName = errorMessage.split(ambColStr)[1].trim();
             errorMessage = i18n._("Ambiguous column name \"%(colName)s\".",
                 {colName});
+            // Note(danielhollas): Added a more helpful text as a separate string
+            // to preserve existing translations.
+            // I18N: This Oh Noes message follows "Ambiguous column name" SQL error
+            // I18N: Keep the space at the beginning of string.
+            errorMessage += " " + i18n._("Multiple tables that you're joining " +
+               "contain a column with that name. To use that column in your query, " +
+               "specify the table of the column. Example: firstTable.\"%(colName)s\"",
+                {colName});
         }
         const unknownColStr = "no such column:";
         const unknownColError = sqliteError.indexOf(unknownColStr) > -1;
--- a/js/output/sql/sql-output.js
+++ b/js/output/sql/sql-output.js
@@ -105,11 +105,9 @@ window.SQLOutput = Backbone.View.extend({
             // Note(danielhollas): Added a more helpful text as a separate string
             // to preserve existing translations.
             // I18N: This Oh Noes message follows "Ambiguous column name" SQL error
-            // I18N: Keep the space at the beginning of string.
             errorMessage += " " + i18n._("Multiple tables that you're joining " +
-               "contain a column with that name. To use that column in your query, " +
-               "specify the table of the column. Example: firstTable.\"%(colName)s\"",
-                {colName});
+               "contain a column with that name. Specify the name of the table that " +
+               "contains the column, using the format \"tableName.columnName\".");
         }
         const unknownColStr = "no such column:";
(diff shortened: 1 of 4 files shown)


js/output/sql/sql-output.js
```diff
@@ -102,6 +102,14 @@ window.SQLOutput = Backbone.View.extend({
             const colName = errorMessage.split(ambColStr)[1].trim();
             errorMessage = i18n._("Ambiguous column name \"%(colName)s\".",
                 {colName});
+            // Note(danielhollas): Added a more helpful text as a separate string
+            // to preserve existing translations.
+            // I18N: This Oh Noes message follows "Ambiguous column name" SQL error
+            // I18N: Keep the space at the beginning of string.
+            errorMessage += " " + i18n._("Multiple tables that you're joining " +
+               "contain a column with that name. To use that column in your query, " +
+               "specify the table of the column. Example: firstTable.\"%(colName)s\"",
```
pamelafox, line 111: Hm, seeing it in the screenshot, the example is throwing me off. It's weird for me to see a generic table name (firstTable) and then a specific column name (id). It's halfway between being an example and being the real thing.
 
Another idea is this sentence instead of "To use..."
"Specify the name of the table that contains the column, using the format "tableName.columnName".
(I'm not sure we need "to use that column in your query", since the error implies they are trying to use it, right?")
What do you think? What's most helpful?

js/output/sql/sql-output.js
```diff
@@ -102,6 +102,14 @@ window.SQLOutput = Backbone.View.extend({
             const colName = errorMessage.split(ambColStr)[1].trim();
             errorMessage = i18n._("Ambiguous column name \"%(colName)s\".",
                 {colName});
+            // Note(danielhollas): Added a more helpful text as a separate string
+            // to preserve existing translations.
+            // I18N: This Oh Noes message follows "Ambiguous column name" SQL error
+            // I18N: Keep the space at the beginning of string.
+            errorMessage += " " + i18n._("Multiple tables that you're joining " +
```
pamelafox, line 109: I'm fine with leaving "Ambiguous column name" as a prefix to the error.

js/output/sql/sql-output.js
danielhollas: Agreed, I was not very satisfied either. I'll change it tomorrow. We do have access to the SQL statement that throws the error so we could try to extract at least one table name with some clever regexing. LMK if you think I should pursue that.

pamelafox: I don't think it's worth it, since it's possible they mean the other table, right?

danielhollas:
Updated wording (and rebased on master)

image

Another option how to help student would be to mention this in the SQL documentation, perhaps for [inner join and outer join](https://www.khanacademy.org/computer-programming/sql-join-on-tables/5409956539006976).

js/output/sql/sql-output.js
danielhollas: @pamelafox right, good point, that would actually perhaps make it more confusing. I updated the wording and rebased on master, see comment below. Thanks!

danielhollas:
@pamelafox I've also fixed the SQL demo page, Oh Noes was unhappy about "char" and "int".

image

pamelafox:
I think the talkie does mention it, per this transcript screenshot?

Screen Shot 2020-04-08 at 3 45 24 PM

PR #730, opened 2020-04-08
Fix and correctly pluralize row/rows DB header.

danielhollas:

It seems that #691, which made the "row/rows" strings translatable, did not work as intended. On the current live site, the header contains only a number, without "rows"

image

I'm not entirely sure what's wrong with the original approach, but I think it's connected to the fact the the `rowsMsg` is accessed inside the `{{each tables}}` context in the Handlebars template. When I moved `{{rowsMsg}}` elsewhere, it did render correctly.

This PR hopefully fixes that and also fixes the pluralization of the i18nized string.

No.

No, I tested visually on the SQL test page. I am nervous though that while testing different approaches, I sometimes broke the SQL output page (it was blank), but the tests were happily passing away. It seems the current tests only check the Oh Noes errors, but not the actual output of the SQL page. I don't know how difficult would it be to add such a test. 

### Risks involved

We need to carefully check that this worked in prod after deploy.

No.

Verify that the table headers show "1 row" or "X rows". We also should verify that the generated Crowdin strings are correct. I will take care of that.

### DIFF
--- a/build/js/live-editor.output_sql.js
+++ b/build/js/live-editor.output_sql.js
@@ -7,78 +7,60 @@ window["Handlebars"]["templates"]["sql-results"] = Handlebars.template({"1":func
     + container.escapeExpression(((helper = (helper = helpers.databaseMsg || (depth0 != null ? depth0.databaseMsg : depth0)) != null ? helper : helpers.helperMissing),(typeof helper === "function" ? helper.call(depth0 != null ? depth0 : (container.nullContext || {}),{"name":"databaseMsg","hash":{},"data":data}) : helper)))
     + "</h1>\n";
 },"3":function(container,depth0,helpers,partials,data) {
-    var stack1, helper, alias1=depth0 != null ? depth0 : (container.nullContext || {});
+    var stack1, helper, alias1=depth0 != null ? depth0 : (container.nullContext || {}), alias2=helpers.helperMissing, alias3="function", alias4=container.escapeExpression;
 
   return "        <table class=\"sql-schema-table\" data-table-name=\""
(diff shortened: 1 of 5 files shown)


pamelafox:
This works for me locally. Thanks for the fix! I wonder if my original one ever worked.

danielhollas:
Great! @pamelafox LMK when you deploy and I can make a post on the forum about the new strings (or you can).

pamelafox:
Deployed (all three)! That'd be great if you could post in the forum. Please thank yourself profusely for the strings.

pamelafox:
(I think you'll need to wait til tomorrow for the strings to show up.)

pamelafox:
Also, thank you profusely for the strings, from me. :-)

danielhollas:
Hehe. I prefer the others to do the thanking. :blush:  

Thanks for all the reviews and deploy and for teaching me the word "profusely". I will be profusely using it from now on. :slightly_smiling_face: 
 
I don't see the strings on Crowdin yet, but I see the "row" header live on Khan! :tada: 

image

pamelafox:
I found the strings!

https://crowdin.com/translate/khanacademy/26342/enus-cs#6636170
https://crowdin.com/translate/khanacademy/26342/enus-cs#6636174
https://crowdin.com/translate/khanacademy/26342/enus-cs#6636172
https://crowdin.com/translate/khanacademy/26342/enus-cs#6636168
https://crowdin.com/translate/khanacademy/26342/enus-cs#6636166

danielhollas:
Nice! And all the strings are properly pluralized and have comments in the context! 🎉

PR #739, opened 2020-09-05
Fix HTML editor resize bug

### LINKED ISSUE #738: Resizing a webpage's width sets it to 400px tall regardless of starting height
When a webpage's height is changed through settings or a query string, then drag-resized by the user, the frame snaps down to 400px height and can't be fixed without reloading the page. A bit of debugging narrows down the issue to https://github.com/Khan/live-editor/blob/master/js/live-editor.js#L1405. Since no height is passed to the updateCanvasSize function when it's resized by the user, it uses the default, which is 400.
```js
height = height || this.defaultOutputHeight;
```
could be changed to
```js
height = height || this.editorHeight || this.defaultOutputHeight;
```
I tested this with the console after the page loaded and it fixed it.

Willard21:
Fixes #738
When the HTML editor is drag-resized by the user, the height always returns to 400. This PR attempts to fix that by setting the default height to the current height if it exists when resizing.

No. This is a very small change.

No. No intended functionality was changed.

### Risks involved
This change shouldn't carry any risks as long as the editor height is always either falsey or a number, or a height is passed to the function. From what I can tell, the only time the height _isn't_ passed to the function is when the user resizes it.

No.

I can't verify that this doesn't break anything on page load because the horizontal resize code appears to be in a different repo, but I tested the updated function in production by pasting it into the Javascript console on any resized webpage program, then resizing it.
```js
window.LiveEditor.prototype.updateCanvasSize = function updateCanvasSize(width, height) {
    width = width || this.defaultOutputWidth;
    height = height || this.editorHeight || this.defaultOutputHeight;

    this.$el.find(this.dom.CANVAS_WRAP).width(width);
    this.$el.find(this.dom.ALL_OUTPUT).height(height);

    // Set the editor height to be the same as the canvas height
    this.$el.find(this.dom.EDITOR).height(this.editorHeight || height);

    this.trigger("canvasSizeUpdated", {
        width: width,
        height: height
    });
}
```

### DIFF
--- a/build/js/live-editor.ui.js
+++ b/build/js/live-editor.ui.js
@@ -2092,7 +2092,7 @@ window.LiveEditor = Backbone.View.extend({
 
     updateCanvasSize: function updateCanvasSize(width, height) {
         width = width || this.defaultOutputWidth;
-        height = height || this.defaultOutputHeight;
+        height = height || this.editorHeight || this.defaultOutputHeight;
 
         this.$el.find(this.dom.CANVAS_WRAP).width(width);
         this.$el.find(this.dom.ALL_OUTPUT).height(height);
--- a/js/live-editor.js
+++ b/js/live-editor.js
@@ -1402,7 +1402,7 @@ window.LiveEditor = Backbone.View.extend({
 
     updateCanvasSize: function(width, height) {
         width = width || this.defaultOutputWidth;
-        height = height || this.defaultOutputHeight;
+        height = height || this.editorHeight || this.defaultOutputHeight;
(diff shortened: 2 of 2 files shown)


Willard21:
Going through the bug tracker, I noticed that the editor height is apparently supposed to be adjustable, and is being prevented by a bug, so maybe defaulting the canvas height to the editor height isn't the ideal fix. Instead, perhaps `defaultOutputHeight` should be set to `options.height || 400` during initialization (and same with width for consistency).

PR #741, opened 2020-09-14
Remove global i18n references and support Node v12

jeresig:
We had a couple references to a global i18n object, but this is easy enough to fix!

Also, in order to get the build running, I had to pin the version of graceful-fs to v4 and use npm-shrinkwrap.json to force the sub-dependencies to use it too.

Issue: none

Test plan:
Hopefully this should "just work" inside Webapp?


pamelafox:
Hm, the tests have a funky error that I haven't seen before. I'll try your workaround locally. I notice other options are upgrading gulp, but I don't know how much havoc that wreaks.
The challenge tests reference global $._(), is that a problem for what you're trying to accomplish?

pamelafox:
@jeresig Can you update your branch now that I've done the gulp upgrade? (You shouldn't need yarn/shrinkwrap/etc)

jeresig:
@pamelafox Done!

pamelafox (approved):

PR #742, opened 2020-09-15
Upgrade gulp to gulp 4

pamelafox:

This PR upgrades gulp to gulp 4. There were issues with gulp 3 and the current node, so this resolves those issues.

I increased the timeout for tests to take (as the first test in the SQL output suite kept failing, despite seemingly being within the timeout), so that could increase time for tests to run. I don't think it practically does, however.

No :)

### Risks involved

There could be some flakiness with the new gulp that we haven't encountered yet. Wonder how it'll do on Travis.

No

npm run build
npm run test


tests/output/sql/index.html
```diff
@@ -22,8 +22,8 @@
         <script src="../../shared_utils.js"></script>
         <script>
             runTestFiles([
-                "output_test.js",
-                "assert_test.js"
+                "assert_test.js",
```
pamelafox, line 25: I rearranged the order to match the webpage order (assert, then output) as one way to encourage mocha not to flake out on the first test.

testutil/gulp-mocha-runner.js
```diff
@@ -78,7 +78,26 @@ var plugin = function (options) {
             }
 
             const chrome = await launchChrome();
-            const protocol = await CDP({port: chrome.port});
+
+            let protocol = null;
```
pamelafox, line 82: This is from: https://github.com/discourse/discourse/blob/master/test/run-qunit.js
Found it on this issue report:
https://github.com/GoogleChrome/chrome-launcher/issues/145

Not sure why we didn't run into it before.

jeresig (approved):
Very cool - thank you for taking the time to dig into this so that we can have something that works as expected!

pamelafox:
Travis CI was happy! :)

PR #756, opened 2022-06-29
[🔥AUDIT🔥] Wait for output frame to be active before beginning execution.

jeresig:
🖍 _This is an audit!_ 🖍

## Summary:
We were waiting for the frame to load, which was happening too early with our recent move to React. We should wait for the frame to tell us we're active.

Issue: FEI-4639

## Test plan:
I imported this into Webapp and loaded up:
http://localhost:8088/computing/computer-programming/html-css-js/html-js-dom-modification/pc/challenge-avatar-attributes

I added in some logging and confirmed that we now wait for the "active" event to come back before we send off code to be executed. In my limited testing so far this seems to be very successful!

### DIFF
--- a/build/js/live-editor.output.js
+++ b/build/js/live-editor.output.js
@@ -490,9 +490,11 @@ window.LiveEditorOutput = Backbone.View.extend({
         var parentWindow = window.parent;
         // Ignore any attempts to send a message to the same window
         if (parentWindow === window) {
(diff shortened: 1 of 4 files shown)


kevinbarabash (commented):
LGTM.  This seems more robust than the previous solution.  It seems plausible that the editor UI should always be ready before the frame finishes loading.

PR #761, opened 2024-12-17
Only set window.KAInfiniteLoopProtect once in webpage-output.js

kevinb-khan:

This fixes an issue where sometimes the browser will report a `TypeError` because `window.KAInfiniteLoopProtect` was readonly.  The reason why it was readonly is because `StateScrubber` locks down all of the properties on `window`.  I'm not sure how the webpage environment was working before this changing (probably a timing thing).

There are no performance implications for this change.

I have not added any new tests for the changes.  It's hard to write an automated test for given live-editor's current testing infrastructure.

### Risks involved

None that I can think of.

No

I verified this by:
- applying these changes to services/static/node_modules/live-editor/build/js/live-editor.output.js directly
- restarting vitejs in hotel
- loading http://localhost:8090/computer-programming/new/webpage
- modifying the HTML code in the editor and seeing the iframe update
- I tested this in Chrome, Safari, and Firefox


jeresig (approved):
Thanks Kevin!

